PR #328, opened 2023-02-23
Technical Debt Payoff

dayhaysoos:
This pull request is to relieve the developer docs site from a lot of the inherited technical debt.

## Big changes that were made

1. Removed a lot of the custom "Swizzle" components used under the theme folder. These components were very complicated and made things harder to manage and update if needed. Most of the components were blog related anyway. I scrapped these components in favor of relying on CSS to target the necessary components.

2. Removed most of the tailwind theme config. I kept some styles around just to make things a little easier, but the previous config overwrote almost everything as far as styling. It's very convoluted and made it hard to update even the smallest styles.

## What this PR is for

I mostly just want a visual test. A comparison between the site that's currently up right now:

https://developer.tbd.website/

and the same site without the technical debt:

https://deploy-preview-328--tbd-website-developer.netlify.app/

Things definitely aren't pixel perfect, but they should be close enough to be acceptable. If you see any huge, glaring issues, please point them out and I will address them ASAP.


ALRubinger (changes requested):
@dayhaysoos wow; this looks really close. I'm impressed! I've gone through a visual review as you've requested and here's what I've caught:

* Home Page all CTA buttons "Read More", "Learn More", etc: link to the right pages, but the positioning of the page is not at the top, more like anchored in the middle somewhere? For instance "Read More" under "Our Philosophy" links to this:

image

* Open Source page: LGTM!

* Project Page: Same issue as described above for the CTA buttons "View Project"; they link correctly but land the user scrolled down some, not at the top of the page.

* Overall note: there are some vertical spacing differences in your branch from the original, and I prefer them. :)

* `/projects` page: Larger font for the section "New contributors should...", closer to as it is on the existing site.

* `/learn` page: some minor spacing differences, but again, LGTM.

* `/blog` page: Same note as above visually. Your dev branch is missing some (half, I think?) blog entries on the live site now; have you accounted to make sure these are carried over after merge?

* `/events` page: Your version has a sidebar which shows a listing of the events; the live site doesn't. Ping @angiejones on desired behaviour? I think having the sidebar and listing is preferable, but she may have more context, so let's check.

* `/docs`: I like that your version keeps the right sidebar in view no matter where you scroll on the page over the version on the live site, which doesn't.

* Footer: The live site has a mouseover effect, a yellow underline, which we've lost in this proposed branch. Also, is there a closer font/size/weight to match the "Legal" and "Code of Conduct" as the live site has?

dayhaysoos:
Thank you for your feedback @ALRubinger @kirahsapong !

## Things I've taken care of:
- Footer font
- Spacing issues on the web5 page
- Removed underline on sidebar
- Showing all blog posts now, not just "most recent"

One thing I wanted to talk about, regarding the feedback on the Component links:
- The way that is set up right now is really weird, I tried putting it back the way it was in the old code base but it was introducing more and more bugs. Since the only things those links are showing are ReadMe's, I decided to link off to those ReadMe's instead.

It doesn't really make sense to have those documents there anyway, considering that the docs section is supposed to have the info that's showing there. Would like to talk this through at some point.

dayhaysoos:
Oh, forgot to mention. On the previous site, there was no need to use `useEffect` to scroll to the top when changing pages, I want to figure out how to fix this without some hacky React trick. If it takes too long, though, I'll just commit to `useEffect`.

PR #338, opened 2023-03-02
Bug: Broken link in blog post, Missing https:// protocol

leabs:
I noticed this link is broken on this blog post, it is missing the https:// protocol before the url which renders the link as a subpage of developer.tbd.website/blog. It might be beneficial to have a link check action on deployment.


dayhaysoos:
Hey @leabs, I appreciate the pull request! It looks like I took care of this with the recent PR I just merged #328.

I do agree, a link check action would be helpful, I think Docusaurus comes with a built-in way to check itself as well. Going to make an issue for it, thank you! Going to close

Issue #387, opened 2023-04-11
Redirected needed for VC Issuance tut/blog post


angiejones:

ALRubinger:
Proposed fix in #393 for @angiejones confirming this is the right fix, else need more context as to where the "old link" is located

ALRubinger:
Bumping to Phase 3

ALRubinger:
Reassigning to @angiejones for Q above, bumping from Phase 3

PR #393, opened 2023-04-14
Update redirect config to new blog location

ALRubinger:
#387

@angiejones does this satisfy what you're requesting? If not, where is the "old" link you're referring to in #387?

Pending @angiejones review, leaving to @dayhaysoos to field the PR and merge alongside the other work flowing in for Phase 2.

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,3 @@
 [[redirects]]
   from = "/learn/tutorials/issue-verifiable-credential-manually"
-  to = "/docs/tutorials/issue-verifiable-credential-manually"
+  to = "/blog/issue-verifiable-credential-manually"


angiejones (commented):
yes but it needs to be merged as part of phase-2 as thats where the changes are

dayhaysoos:
We should probably add a Redirect on Netlify config just in case anyone tries to access the old link, yeah? If I remember correctly, this blog was announced and shared publicly

dayhaysoos:
Oh just found there was an issue for this #387, going to check now

dayhaysoos (approved):

dayhaysoos:
I'm going to make sure this gets into main

angiejones (commented):
from should be: /docs/tutorials/issue-verifiable-credential-manually/

Issue #397, opened 2023-04-14
Lock version of Web5 for Phase 2


ALRubinger:
From Discord:
https://discord.com/channels/937858703112155166/969272658501976117/1096373020093861949

What's the version of Web5 we're using for Phase 2? And what's the direct link we'll advise to pull it from the CDN?

Please assign to me when answered? I'll update the Quickstart, for instance:

```
import { Web5 } from 'https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.3.1/dist/browser.min.js'
```

I'll also update Devin's example: https://github.com/TBD54566975/web5-quickstart-widgets/blob/main/index.html#L172

ALRubinger:
As decided w/ @csuwildcat:

* We'll continue with phase-2 publish today, merging to main when all blockers are cleared
* No splashy announcement about it on Discord or elsewhere quite yet
* OSE will get us a version number to lock docs against - that will resolve @ALR | TBD blocker https://github.com/TBD54566975/developer.tbd.website/issues/397
* OSE is working on a set of API changes to land early next week. 
* So we'll fast follow with a docs update to upgrade the version of web5-js that we're documenting, and also update the docs to match.
* That process will actually be really helpful - it'll show everyone on the team all the things that need to get updated when we push new versions.

ALRubinger:
Surfacing what likely needs to change when we update the version:

https://github.com/TBD54566975/developer.tbd.website/issues/392#issuecomment-1509238801

There may be API changes in the new version, and we'll have to update `web5-quickstart-widgets`, make sure that's working, then bring that code into the Quickstart.

No updating Quickstart code from anything that's not verified as working in the `web5-quickstart-widgets repo.

ALRubinger:
@mistermoe cut `0.4.0`!

For NPM: `npm install @tbd54566975/web5@0.4.0`
For CDN: `import { Web5 } from 'https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.4.0/dist/browser.mjs';`

This resolves the OSE dependency and now what's left in this issue is updating examples and docs to match the `0.4.0` upgrade.

ALRubinger:
Confirmed this is working in `web5-quickstart-widgets` as a drop-in upgrade w/ no code changes necessary: 

https://github.com/TBD54566975/web5-quickstart-widgets/issues/1

ALRubinger:
[...]

PR #403, opened 2023-04-14
Issue #397 Lock Quickstart for phase-2 to web-js 0.4.0

ALRubinger:

PR #405, opened 2023-04-17
Lock 0.4.0

dayhaysoos:
This addresses issue #397 

I already tested locally and it worked just fine (including build command)


ALRubinger:
Where did my commit 051deb9 go? PR #403? Let's use that.

As is on `main` and deployed the Quickstart is not using the right versions.

ALRubinger:
Resolved in f2ac3102dc4ebe2f534c9adf3dbb45d7a4be5e5c

PR #414, opened 2023-04-19
Remove custom code block font size for tokens of type 'plain'

kirahsapong:
This PR addresses font size of code block strings with token type "plain"


angiejones (approved):
bless you @kirahsapong!

PR #422, opened 2023-04-25
DRAFT: Upgraded playground blog post

andorsk:
A draft for an upgraded playground blog post. Sharing mostly so there's no overlapping efforts. This DRAFT is not finalized yet, and we are still reviewing it on our side.


dayhaysoos:
Hey @andorsk ! Do you mind pointing this PR to the `phase-3` branch for now? I'm sure this change would be fine on main as is, but right now we're treating `phase-3` as the staging branch for the work we're pushing this week and it would prob be safer to go about things this way just in case.

Lemme know if you have any issues with this, I'm happy to help however I can!

andorsk:
@dayhaysoos yea! no problem. Can do!

andorsk:
@dayhaysoos moved over. See #423

dayhaysoos:
🫡🫡🫡🫡

PR #429, opened 2023-04-27
Build Folder and Record in Quickstart Updates 0.6.0

bobbilee19:
Went ahead and found all places where code and copy needed to be updated to 0.6.0 (inline with Quickstart changes) and myRecord > record in Quickstart updated well.

One thing we didn't change yet in this PR was the Build > Apps > ToDo app because it's not a breaking change right now bc the app example pinned at the top is using 0.4.0 SDK.

### DIFF
--- a/docs/web5/build/decentralized-identifiers/how-to-use-external-dids.mdx
+++ b/docs/web5/build/decentralized-identifiers/how-to-use-external-dids.mdx
@@ -11,12 +11,15 @@ const web5 = new Web5();
 const myDid = await web5.did.create('ion');
 
 // Save your newly created DID in your DWN
-web5.did.register({
-      connected: true,
-      did: myDid.id,
-      endpoint: 'app://dwn', 
-      keys: myDid.keys[0].keypair
-    });
+await web5.did.manager.set(myDid.id, {
+    connected: true,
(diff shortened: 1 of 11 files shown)


docs/web5/build/decentralized-identifiers/how-to-use-external-dids.mdx
```diff
@@ -52,35 +55,4 @@ In order to bring an external DID into your `Web5` project, simply replace the `
...
-      connected: true,
-      did: myDidSSI.didDocument.id,
-      endpoint: 'app://dwn',
-      keys: keypair
-    });
-```
-
-As long as the `ssi-service` and `Web5` instances that are creating the keys live in the same application, this is perfectly safe to do. If, however, you have a separate application running `ssi-service` that generates a `did`, export that `did` to plain text or some other more portable format, and attempt to import that `did` document into your app, you’d be doing so in an unsafe manner.
\ No newline at end of file
+```
```
bobbilee19, line 58: We removed this for now because we'd like to come back to it. SSI SDK is not instantiated before being referenced so we'd like to open this up for discussion.

ALRubinger:
Added some more fixes to Quickstart, squashed, put upstream in 1bdde5f3822fafc4dbd8daa990bc1b794210404d

PR #444, opened 2023-05-05
updated ssi service playground image

andorsk:
updated the SSI service playground image. Checked in beforehand and should be ok, but LMK if you have any questions.


bobbilee19:
LGTM!

andorsk:
thank you!

PR #491, opened 2023-05-21
WIP: Removal of DID example and on-chain reference

bobbilee19:


ALRubinger (approved):
LGTM for now, removing reference to VCs being on chain. @bobbilee19 revisit when ready in new issue and PR

PR #557, opened 2023-06-14
update protocols example

chris-tbd:
Fix protocols example code

### DIFF
--- a/docs/web5/learn/protocols.md
+++ b/docs/web5/learn/protocols.md
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
 
 ## Protocols in Practice
 
-To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library, where `myDid` is your DID object and `protocolObject` is the protocol json defined in the previous section:
+To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library:
 
 ```js
-const response = await web5.dwn.protocols.configure(myDid.id, protocolObject);
+const { protocol } = await web5.dwn.protocols.configure({
+  message: {
+    definition: {
+      protocol: "https://photos.org/protocol",
+      types: {
+        album: {
+          schema: "https://photos.org/protocol/album",
+          dataFormat: ["application/json"],
+        },
+        photo: {
+          schema: "https://photos.org/protocols/photo",
+          dataFormat: ["application/json"],
+        },
+        binaryImage: {
+          dataFormat: ["image/png", "jpeg", "gif"],
+        },
+      },
+      structure: {
+        album: {
+          $actions: [
+            {
+              who: "recipient",
+              can: "read",
+            },
+          ],
+        },
+        photo: {
+          $actions: [
+            {
+              who: "recipient",
+              can: "read",
+            },
+          ],
+          binaryImage: {
+            $actions: [
+              {
+                who: "author",
+                of: "photo",
+                can: "write",
+              },
+            ],
+          },
+        },
+      },
+    },
+  },
+});
+
+protocol.send(myDid); // sends the protocol configuration to the user's other DWeb Nodes.
 ```
 
 Once you’ve installed that protocol to your app, you’re ready to communicate using the schema and permissions it defines.


docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
 ```js
-const response = await web5.dwn.protocols.configure(myDid.id, protocolObject);
+const { protocol } = await web5.dwn.protocols.configure({
+  message: {
+    definition: {
+      protocol: "https://photos.org/protocol",
+      types: {
+        album: {
+          schema: "https://photos.org/protocol/album",
+          dataFormat: ["application/json"],
```
frankhinek, line 307:
```suggestion
          dataFormats: ["application/json"],
```

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+        },
+        photo: {
+          schema: "https://photos.org/protocols/photo",
+          dataFormat: ["application/json"],
```
frankhinek, line 311:
```suggestion
          dataFormats: ["application/json"],
```

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+        },
+        binaryImage: {
+          dataFormat: ["image/png", "jpeg", "gif"],
```
frankhinek, line 314:
```suggestion
          dataFormats: ["image/png", "jpeg", "gif"],
```

frankhinek (changes requested):

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
 
 ## Protocols in Practice
 
-To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library, where `myDid` is your DID object and `protocolObject` is the protocol json defined in the previous section:
+To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library:
```
angiejones, line 297: change 'install that protocol to your app' to 'install that protocol to your DWN'

angiejones (changes requested):

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
 
 ## Protocols in Practice
 
-To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library, where `myDid` is your DID object and `protocolObject` is the protocol json defined in the previous section:
+To use a protocol in your app, you’ll need to install that protocol to your DWN. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library:
 
 ```js
-const response = await web5.dwn.protocols.configure(myDid.id, protocolObject);
+const { protocol } = await web5.dwn.protocols.configure({
```
jiyoontbd, line 300: isn't there a protocol _definition_ defined in the `Defining Protocol` section on line 48? could we just reference that as a variable of some kind here? 

also, the protocol that's written here talks about albums, photos, and binary image types...

but the protocol defined above (and the premise of this tutorial) discusses message, reply, image, and caption types, so i think referencing an entirely new protocol halfway in the tutorial may confuse the reader.

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+              },
+            ],
+          },
+        },
+      },
+    },
+  },
+});
+
+protocol.send(myDid); // sends the protocol configuration to the user's other DWeb Nodes.
```
jiyoontbd, line 348: i don't think `protocol.send` is a real method?

docs/web5/learn/protocols.md
jiyoontbd, line 300: also, `await web5.protocols.configure(...)` returns ProtocolsConfigureResponse which includes `status` and optionally `protocol`, from what i can see in the doc...

```javascript
export type ProtocolsConfigureResponse = {
  status: MessageReply['status'];
  protocol?: Protocol;
}
```

@diehuxx pls feel free to correct me if i'm wrong

jiyoontbd (changes requested):

docs/web5/learn/protocols.md
diehuxx: Yep, this configures the protocol on a dwn, through from this small snippet, I can't tell if the default DWN is the same as the DWN for `myDid`. I'm reading through the rest of the tutorial now. Then `protocol.send()` below also configures the protocol for the DWN of `myDid`. @chris-tbd Are you trying to configure the protocol on two different DWNs? If not, I think you are doing the same thing twice, in two different ways.

diehuxx: It may be useful to get `status` from the response, to show the developer that they could do that if they want to. But Chris, I'll leave it up to you if that's what you want to communicate or if it's clear from other parts of the tutorial.

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+        },
+      },
+      structure: {
+        album: {
+          $actions: [
+            {
+              who: "recipient",
+              can: "read",
+            },
+          ],
```
diehuxx, line 324: Recipients of a message can always read the message, so you don't need this. Also, did this code run for you? It does not work for me, because `who: 'recipient'` must have a corresponding `of` field.

docs/web5/learn/protocols.md
```diff
@@ -294,10 +294,58 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+        },
+        photo: {
+          $actions: [
+            {
+              who: "recipient",
+              can: "read",
+            },
+          ],
```
diehuxx, line 332: Omit here as well

docs/web5/learn/protocols.md
diehuxx: Repeating what I mentioned above. `protocol.send()` is a real method, but I don't think it's what we want here. `web5.dwn.protocols.configure()` already configures the protocol. `protocol.send` also configures the protocol. Unless I'm misunderstanding the intention, we don't need to do it twice.

docs/web5/learn/protocols.md
diehuxx: +1. If we're going for complete copy-and-pasteability, we should also include `const { web5, did: myDid } = await Web5.connect();` so the read knows where we got `web5` and `myDid` from.

If we're going for brevity/skimmability, we should reference the protocol object earlier in the tutorial.

docs/web5/learn/protocols.md
angiejones: we should also document above this the code snippet that recipients can always read

docs/web5/learn/protocols.md
angiejones: I thought `protocol.send()` replicates the protocol across the user's other DWNs. like `records.send()`. No?

jiyoontbd: could you point me to the documentation for `protocols.send()` ? i don't see it in the dwn-api.ts under `protocols` - the two methods i see are `configure` and `query`. i'm using web5-js version `0.7.9`...

jiyoontbd: oh! you know what? i'm confusing `protocols` methods and `protocol` methods. `protocol.send()` real :D

diehuxx: Is the intention in this tutorial that the user has set up multiple DWNs at this point?

Reading through the code in web5-js, you're right that `protocol.send()` does try to send a message to one of the users nodes. But, it doesn't send it to all their nodes. It looks through the DID document to find publicly addressable nodes, and tries sending the message to each node until one succeeds. If the first node to succeed is the local node in the local instance of `web5`, then the other nodes aren't sent the message.

Either way, IIRC the sync mechanism should handle replicating data across a user's nodes automatically. We shouldn't need to re-send the message to a remote node. If sync is working as I thought it was, we should be able to trigger DWN sync and the DWNs will handle it themselves.

diehuxx: @jiyoontbd ope lol sorry for the confusion. DWN methods vs javascript methods. Not a real DWN method, but yes a real javascript method.

chris-tbd: I think this got nixed with sync

docs/web5/learn/protocols.md
chris-tbd: Same as above

docs/web5/learn/protocols.md
```diff
@@ -294,41 +294,41 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+    message: {
+      definition: protocolDefinition 
+    }
+});
 ```
 
 Once you’ve installed that protocol to your app, you’re ready to communicate using the schema and permissions it defines.
 
-Building on our social media example, let’s say that you wanted to post a message to your friend Alice. First, you can check if she also has the `social-media` protocol installed by running:
+Building on our social media example, let’s say that you wanted to post a message to your friend Alice. First, ensure that she also has the `social-media` protocol installed on her DWN. Then, you can write to her DWN via the `social-media` protocol using:
```
jiyoontbd, line 309: i think the protocol would be `http://social-media.xyz`

docs/web5/learn/protocols.md
```diff
@@ -294,41 +294,41 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
...
+await web5.dwn.records.create({
   data: 'Hello, world!',
   message: {
-    recipient: 'did:example:alice'
-    schema: 'message',
+    recipient: aliceDid,
+    schema: 'http://messageschema',
     dataFormat: 'text/plain',
-    protocol: 'social-media',
+    protocol: protocolDefinition.protocol, // or, "http://social-media.xyz", and not 'social-media'
```
jiyoontbd, line 319: probably good to get rid of the comment `// or, "http://social-media.xyz", and not 'social-media'` here now :)

docs/web5/learn/protocols.md
```diff
@@ -294,41 +294,41 @@ Additionally, you’ll notice the `image` object below it also defines `actions`
 
 ## Protocols in Practice
 
-To use a protocol in your app, you’ll need to install that protocol to your app. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library, where `myDid` is your DID object and `protocolObject` is the protocol json defined in the previous section:
+To use a protocol in your app, you’ll need to install that protocol to your DWN. You can do so using the following snippet that leverages our [web5.js](https://github.com/TBD54566975/web5-js) library, where `protocolObject` is the messaging protocol object from above:
```
jiyoontbd, line 297: i think this variable name should be called `protocolDefinition` because we use that variable name in the below code snippet

jiyoontbd (approved):

angiejones (approved):

Issue #538, opened 2023-06-06
Update Quickstart: Instantiation


angiejones:
https://developer.tbd.website/docs/web5/quickstart/#1-instantiate-web5

Need to update the description of the title and description of Step 1: Instantiate Web5

It now instantiates Web5 and also provides a class instance representing the decentralized identifier that was created or attained connection to

bobbilee19:
https://github.com/TBD54566975/developer.tbd.website/issues/538

Issue #532, opened 2023-05-31
Unify all docs on developer.tbd.website to respect one dependency


ALRubinger:
Introduce a Bill of Materials. 

This will give us a single place to update, for instance, a Web5 or SSI dependency. All downstream efforts with the Web5 dependency: TODO App Guide, TODO app starter template, TODO app finished example, Quickstart, Quickstart index.js, Quickstart front-end widget - respect and use that one declared dependency.  This gives us one place to update across the entire docs suite - no more cascading updates in several places and issuing new releases across different repos.

@dayhaysoos to dig in and provide an RFC for stakeholder review.

dayhaysoos:
**Proposal: Centralization of Examples in developer.tbd.website Repo**


First of all, I just want to say that this is one step of many towards making a smoother process when it comes to going from release to docs being updated.

I suggest we add a dedicated `examples` folder in the `developer.tbd.website` repository to consolidate the separate repositories we currently maintain for examples.

Currently, the following examples are maintained in separate repositories:

- TODO App Guide
- TODO App starter template
- TODO app finished examples
- Quickstart frontend Widget

Managing these separate repositories has been cumbersome and lead to inconsistencies and tracking issues. I think a good first step would be to move all these repos under an `examples` folder. By making this folder, it would help with:

- **Improved Efficiency**: Having a centralized location for all examples will streamline our update process and eliminate the need to switch between repositories.
- **Ease of Tracking**: We can track changes and updates across all examples more effectively when they are in one location.
- **Consistency**: It will help to ensure that changes made are consistent across all examples.

I also propose that we should make it a requirement that each example should work "out of the box". If the example is in an index.html, we should be able to see it working from copying/pasting into a local html file, for example This would help avoid any path or dependency issues and improve the experience for developers who are trying out or learning from our examples.

The scope of this step is pretty small, but I would love to talk about expanding from here. Here are some things that we should talk about:

- Reactifying the Quickstart to make it easier to reason about and edit
- Ways to synchronize between examples and tutorials (are there certain tools we can use to handle this, ways we can template it?)
- Editing the examples/tutorials so they are 1:1 copies of each other (basically, the same exact code being written in the example is also being used for the tutorial, I noticed we had to refactor things to fit the appropriate context)


I'll leave it here for now, feel free to share any thoughts or concerns around this task.

bobbilee19:
1000% yes. Thank you for writing this all out. And apologies I didn't see it right away, so sorry.

[...]

PR #564, opened 2023-06-22
Pnpm

dayhaysoos:


ALRubinger (changes requested):
PNPM seems to be working great! Ran through it and w/ @dayhaysoos came across changes to fulfill the underlying #532 issue.

- [x] Right now - the `pnpm` arrangement doesn’t yet fulfill the issue #532 to declare and respect one dependency (for instance Web5) across all examples and the site. For instance, “examples/todo-app/package.json” has “"@tbd54566975/web5": “0.7.8”; expecting this to be declared once uniformly at the top level so all examples and the site itself use the same version. That gives us one place to update as defined by #538.
- [x] Similar to above, Dev Site is using its own declaration of Web5 in its package.json: "@tbd54566975/web5": "0.7.9" 
- [x] Figure a way to inline the declared dependency into the dev site QuickStart instructions:
Pasted Graphic 1

- [ ] Web5-quickstart-widgets index.html is still using a direct dependency declaration for Web5, outside the new pnpm system: “import { Web5 } from '[https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs’](https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs%E2%80%99);”
- [x] Renames suggested: todo-app > todo-completed, and todo-app-tutorial > todo-starter
- [x] Suggest script renames: “start:todo-tutorial” > “start:tutorial-todo-starter”, “start:filter-todo” > “start:tutorial-todo-completed”
- [ ] Todo-app-tutorial - needs updating to get the starter app (web5-tutorials) and completed example (incubating-web5-labs) from their new locations (under the monorepo in examples folder)
- [x] pnpm-lock.yaml to .gitignore
- [ ] tbd-project-template stuff - license, CoC, Governance, etc.
- [x] In “site commands”, prefix the commands with “pnpm”
- [x] In site commands, “pnpm start, not start:site”
- [ ] On http://localhost:3000/docs/apis/ssi-service, templating looks borked:
Pasted Graphic

- [ ] “Running commands” section in README: what does “target specific packages” mean? Maybe this section could have an example to show what you’d be doing with this command, for instance updating web5?
- [ ] Suggest renaming “docs” package to “site” to match the folder structure and refer to the site that we’re building.
- [ ] `npm` older commands like "install" still runnable directly, but can lead to the wrong build outputs? Can we remove `npm` from working at all, signifying that `pnpm` should be the build system of record?

dayhaysoos:
> Right now - the pnpm arrangement doesn’t yet fulfill the issue https://github.com/TBD54566975/developer.tbd.website/issues/532 to declare and respect one dependency (for instance Web5) across all examples and the site. For instance, “examples/todo-app/package.json” has “"@tbd54566975/web5": “0.7.8”; expecting this to be declared once uniformly at the top level so all examples and the site itself use the same version. That gives us one place to update as defined by https://github.com/TBD54566975/developer.tbd.website/issues/538.

We have a dependency at the root package.json now that says to use web5 0.7.10, however, that doesn't actually install the required package inside each project. It allows us to locally test versions. 

I made a specific command for updating all dependencies in examples to take care of this:

`pnpm update:web5" will recusrively add the latest version of web5 to every example.

Here is what the package.json looks like now.

I think this is the best we can do for now, I'm sure we can get some custom hooks going to automate this just a bit further.

ALRubinger:
Got off a pair exercise w/ @dayhaysoos - here's where we landed:

* He's put the `web5` dependency into the top level `pnpm` 
* That makes for one declared dependency for `web5`, used by the site and all examples/demos/starter templates/etc
* Additional dependencies that are specific to the site or examples/demos/starters/etc can still go in the `package.json` of that submodule

So with that arrangement, let's say @bobbilee19 wants to upgrade Web5. She could either:

1) Update the version number in the top-level `package.json` once, or
2) Run a `pnpm` command at the root level to have `pnpm` do the above (@dayhaysoos is documenting how in the README)

@bobbilee19 - this one is of particular interest: we advise taking the "completed" `index.js` in the Quickstart and instead of it being a free-floating file the user downloads, make it into a simple "quickstart-completed" application mirroring how we do it in the completed TODO app. That way:

* A user has the fully runnable application available to them, which includes the `package.json` needed to run it
* We have a uniform approach to completed applications and we can upgrade and test them all in the same way; no more standalone floating Node scripts to independently patch and manually check when we upgrade

@dayhaysoos is working through the bullets in my PR review above, looks like making great progress. 

And I'm super excited because in addition to easing the maintenance burden, we're giving some really good unifying structure to the examples/completed apps/playgrounds etc which is going to make it really clear to the user how to get them, how to run them, and for us to make testsuites for them.

bobbilee19:
@dayhaysoos oh 100%, consistency, consistency and the two points for why we should do that, 100% agree...for the index.js question above.

dayhaysoos:
>  Figure a way to inline the declared dependency into the dev site QuickStart instructions:

Due to the nature of docusaurus and MDX, I'm using an environment variable to declare the version for Web5 but also adding that value to Docusaurus's Context. We can't simply inject the version where we want because that's not how MDX works, we have to create React components to do it. 

I also think we should get away from declaring which version gets installed for the `npm install` snippets. Probably something we should ask OSE about because this depends on how stable the next releases will be until we get to 1.0.0.

If we should keep the versions inlined on the npm lines, then I'll have to make a component explicitly for that use case in the meantime that takes advantage of the Docusaurus Context.

dayhaysoos:
> pnpm-lock.yaml to .gitignore

I think we actually need the lock file. This guarantees that everyone will be running the same versions of things when they pull. Tbh I can't remember the details but I was taught it's a bad practice to hide lock files.

ALRubinger:
> > pnpm-lock.yaml to .gitignore
> 
> I think we actually need the lock file. This guarantees that everyone will be running the same versions of things when they pull. Tbh I can't remember the details but I was taught it's a bad practice to hide lock files.

You're totally right, and I suppose just needs to be updated/synced alongside `package.json` so when users run the build they don't see diffs.

ALRubinger:
- [ ] Reorganize `examples` folder as @angiejones suggests: https://discord.com/channels/937858703112155166/1111754777307660388/1123013349198794762
- [ ] Update `README` so it's self-serve, reflecting new script changes
- [ ] Existing docs reference `incubating-web5-labs` and `web5-tutorials`, which are now moved to this monorepo

dayhaysoos:
> Web5-quickstart-widgets index.html is still using a direct dependency declaration for Web5, outside the new pnpm system: “import { Web5 } from '[https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs’](https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs%E2%80%99);”

Is this the only index.html example we ever plan on doing? If so, we should leave it as is. Perhaps this will be the only thing we have to take care of manually. Getting this to work within the `pnpm` system is becoming a pain due to CORS related issues. Maybe I'm going about it wrong but I'd like to come back to this after other things are taken care of

dayhaysoos:
The "borked" values are actually coming from the swagger.yaml source:

https://raw.githubusercontent.com/TBD54566975/ssi-service/main/doc/swagger.yaml

Not sure who to contact for this one, any ideas @ALRubinger ?

ALRubinger:
> > Web5-quickstart-widgets index.html is still using a direct dependency declaration for Web5, outside the new pnpm system: “import { Web5 } from '[https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs’](https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs%E2%80%99);”
> 
> Is this the only index.html example we ever plan on doing? If so, we should leave it as is. Perhaps this will be the only thing we have to take care of manually. Getting this to work within the `pnpm` system is becoming a pain due to CORS related issues. Maybe I'm going about it wrong but I'd like to come back to this after other things are taken care of

Based on the other examples I've seen throughout the DWN/Web5 repos, nope. I bet that these declarations are a common pattern for example apps because they're a class of web applicatioj that's not Node-based; they're all in-browser and really show the ease of use of Web5.

ALRubinger:
> The "borked" values are actually coming from the swagger.yaml source:
> 
> https://raw.githubusercontent.com/TBD54566975/ssi-service/main/doc/swagger.yaml
> 
> Not sure who to contact for this one, any ideas @ALRubinger ?

@decentralgabe will know!

And I think we should get this external dependency out of the build, or at least lock it to a version. Changes to `main` at different points in time makes for non-reproducible builds of the dev site. This means that if you run the dev site build on a commit hash X, depending upon when you do that, you can get different results. A build should be "reproducible" by nature of a commit hash build always making the same result.

As a separate and related issue, the external dependency means that if this file is moved/changed, it can block builds of the dev site from completing successfully until corrected. So I recommend bringing these external pieces inside the monorepo.

README.md
```diff
@@ -0,0 +1,76 @@
...
+1. Open a terminal or command prompt.
+2. Run the following command:
+
+```shell
+npm install -g pnpm
+```
+
+3. Wait for the installation process to complete.
+
+> Note: If you already have `pnpm` installed globally, you can skip this step.
```
bobbilee19, line 25: maybe add this at the top of the step?

decentralgabe:
@dayhaysoos which values are borked from the swagger? cc: @andresuribe87

dayhaysoos:
> Based on the other examples I've seen throughout the DWN/Web5 repos, nope. I bet that these declarations are a common pattern for example apps because they're a class of web applicatioj that's not Node-based; they're all in-browser and really show the ease of use of Web5.

Just took a look and I think I know what to do now, going to look at it tomorrow morning.

bobbilee19:
Two places to come back to when we figure out degit...
- Changing where the example / todo-app-starter will go and how we want them to use it (cloning our entire dev docs is a lot for just the starter)... https://github.com/TBD54566975/developer.tbd.website/blob/main/site/docs/web5/build/apps/todo-app-tutorial.mdx#L14
- Changing how we'll link the todo-app-completed...https://github.com/TBD54566975/developer.tbd.website/blob/main/site/docs/web5/build/apps/todo-app-tutorial.mdx#L289

dayhaysoos:
@decentralgabe I don't think this is all of them but from what I can tell, the places where you can see "{{.Something}}" like in this screenshot:

https://user-images.githubusercontent.com/199891/248892883-919d9341-5cda-4198-bcc9-b9b55160e5c8.png

I ran a quick search on "{{" and saw these:

- host: '{{.Server.APIHost}}'
- than {{.Services.CredentialConfig.BatchCreateMaxItems}} items.
- {{.Services.DIDConfig.BatchCreateMaxItems}} items.
- description: '{{.Desc}}'
- version: '{{.SVN}}'

decentralgabe:
@dayhaysoos the problem is those values are populated with a running SSI Service. Without a running service we won't know what to put there (e.g. what service URL, version, etc).

An option may be to have a piece of middleware that detects the pattern with a regex and removes the properties from the resulting yaml file

dayhaysoos:
@decentralgabe I can look into something like that, I guess I'm curious about why these values didn't show up on docusaurus before. Do you think it'd be helpful to look at previous versions of this file for reference?

decentralgabe:
@dayhaysoos we made a change to have these values load dynamically. they weren't being set properly before. the service needs to set these dynamically, so I only can suggest a way to workaround and delete the problematic properties

dayhaysoos:
> And I think we should get this external dependency out of the build, or at least lock it to a version. Changes to main at different points in time makes for non-reproducible builds of the dev site. This means that if you run the dev site build on a commit hash X, depending upon when you do that, you can get different results. A build should be "reproducible" by nature of a commit hash build always making the same result.

As a separate and related issue, the external dependency means that if this file is moved/changed, it can block builds of the dev site from completing successfully until corrected. So I recommend bringing these external pieces inside the monorepo.

I agree with this, I think that maybe this should be another separate issue.

dayhaysoos:
Okay thank you @decentralgabe, I have to do some research because the way we display this is entirely reliant on an external library. Not sure how much control I have over how things are surfaced but I'm sure I can figure something out

dayhaysoos:
Don't think we'll be able to land this today, or we can land this and push some of the tasks to another "sprint". Here's a list of what's left and what's to be said about it:

>  On http://localhost:3000/docs/apis/ssi-service, templating looks borked:
External dependency issue, gabe gave some suggestions on how we can go about this. Just need to be researched.


> Todo-app-tutorial - needs updating to get the starter app (web5-tutorials) and completed example (incubating-web5-labs) from their new locations (under the monorepo in examples folder)

I'ma be honest I don't even know what this means, gonna get clarity from ALR later lol

> Web5-quickstart-widgets index.html is still using a direct dependency declaration for Web5, outside the new pnpm system: “import { Web5 } from '[https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs’](https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs%E2%80%99);”

This might actually be done by EoD today

>  “Running commands” section in README: what does “target specific packages” mean? Maybe this section could have an example to show what you’d be doing with this command, for instance updating web5? 
This is a copy pass, done by EoD today

> npm older commands like "install" still runnable directly, but can lead to the wrong build outputs? Can we remove npm from working at all, signifying that pnpm should be the build system of record?

This is possible, but I'm hesitant on it. I'd like for people to use `npm` if they want for running `pnpm` scripts detailed in the script tags. Just worth having a small discussion about this

ALRubinger:
> > Todo-app-tutorial - needs updating to get the starter app (web5-tutorials) and completed example (incubating-web5-labs) from their new locations (under the monorepo in examples folder)
> 
> I'ma be honest I don't even know what this means, gonna get clarity from ALR later lol

This refers to the docs for TODO App; they still instruct the user to attain the starter and completed example from the old locations: 

- [ ] `git clone https://github.com/TBD54566975/web5-tutorials.git` in the code block at the top
- [ ] `https://github.com/TBD54566975/incubating-web5-labs/tree/0.2.0/todo-app` is the link in "finished version of the ToDo app here" at the end
- [ ] "See it deployed here" at the end points to `https://unrivaled-crumble-56ce70.netlify.app/`, which is configured on Netlify still using the old location `https://github.com/TBD54566975/incubating-web5-labs`

ALRubinger:
> @dayhaysoos we made a change to have these values load dynamically. they weren't being set properly before. the service needs to set these dynamically, so I only can suggest a way to workaround and delete the problematic properties

@dayhaysoos, let's spin this off into a new issue. It's not related to the `pnpm` work here and exists independently. So let's not have it block this PR and address separately.

We need to define a contract by which OSE is providing this source and the site can transpile it into the published version. If OSE folks need to change something about the format they're providing, or move the source file, that's cool; but it should be:

* Explicitly coming to the site via a locked version so nothing changes under the hood inbetween site builds (to my point about [build reproducibility above](https://github.com/TBD54566975/developer.tbd.website/pull/564#issuecomment-1613847833))
* Adhering to the same contract and verified the new docs version looks right before the PR into the dev site is approved and merged to `main`

ALRubinger:
>> Web5-quickstart-widgets index.html is still using a direct dependency declaration for Web5, outside the new pnpm system: “import { Web5 } from '[https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs’](https://cdn.jsdelivr.net/npm/@tbd54566975/web5@0.7.0/dist/browser.mjs%E2%80%99);”

> This might actually be done by EoD today

@dayhaysoos and I agreed:

- [ ] Let's spin this off into a new issue to not block this PR. Updating Web5, until this is done, will have to be done in two locations: the monorepo declaration and for this `index.html` file of the quickstart-widget. Some pre-build step to inject the version into the widget is something Nick will look at as part of the build. But that won't hold up this PR and we'll document all the places to update Web5 in the `README` in the meantime.

ALRubinger:
> This is possible, but I'm hesitant on it. I'd like for people to use `npm` if they want for running `pnpm` scripts detailed in the script tags. Just worth having a small discussion about this

I've learned more about `pnpm` over the past few days in my own building with it - and you're right here. Convention is that it's a drop-in replacement and I've seen most projects simply have in documentation to run `pnpm` and there's no explicit barrier to their using `npm`. We only need to support the documented build flows.

ALRubinger:
- [ ] Update TODO App docs to show user how to get the starter example from the new location (the code block for `git clone` etc)
- [ ] Update TODO App docs to show user how to get completed example from new location
- [ ] Update TODO App Deployment config in Netlify to point to new location in monorepo
- [ ] Remove TODO App from web5-tutorials 
- [ ] Remove TODO App from incubating-web5-labs
- [ ] Quickstart `index.js` should move to monorepo as a `-completed` appl just like we do for TODO App
- [ ] ...And update user instructions to show user how to grab the completed example from new location and run it, just like we do for TODO App
- [ ] Quickstart docs update to show user how to build from new `-completed`

dayhaysoos:
>  Let's spin this off into a new issue to not block this PR. Updating Web5, until this is done, will have to be done in two locations: the monorepo declaration and for this index.html file of the quickstart-widget. Some pre-build step to inject the version into the widget is something Nick will look at as part of the build. But that won't hold up this PR and we'll document all the places to update Web5 in the README in the meantime.

Don't need to make a new issue for this anymore because I think it's working now. Will have to double check but this seems to be done.

dayhaysoos:
> "See it deployed here" at the end points to https://unrivaled-crumble-56ce70.netlify.app/, which is configured on Netlify still using the old location https://github.com/TBD54566975/incubating-web5-labs

Having some trouble getting our netlify set up right to deploy multiple sites. I think for now I'd like to make this an issue and ask for support from Netlify just so we can land this PR today.

Until then, probably won't be a good idea to remove the apps from the other repos. We'll leave it up for people to see in the mean time.

PR #574, opened 2023-06-28
Update 2023-01-04-manually-issue-vcs.md

ALRubinger:
As-is, this link opens as https://json-schema.org/draft/2020-12/schema%22 and 404s

@bobbilee19


ALRubinger:
Alternatively, can make this monospaced using `backticks` if you don't want this to be a link at all

PR #593, opened 2023-07-12
Chatgpt powered search

michaelneale:
chatgpt powered search (similar to chatGpT plugin). 

step 1: ask a question: 

Screenshot 2023-07-13 at 6 40 49 pm

step 2: it thinks about it and consults the same knowledge as the chatgpt plugin: 

Screenshot 2023-07-13 at 6 41 01 pm

Then you get an answer (if on topic, I try to make it not just make things up): 

Screenshot 2023-07-13 at 6 41 12 pm


michaelneale:
@angiejones nearly there - when it is ready to play with will let you know.

angiejones:
> @angiejones nearly there - when it is ready to play with will let you know.

I'm so excited!!!

michaelneale:
cc @kirahsapong halp! You can see the component and where I am placing it, but I can't work out how to get it over the top of the header

site/src/components/ChatGPTSearch.jsx
```diff
@@ -0,0 +1,73 @@
...
+  const handleClose = () => {
+    setIsOpen(false);
+  }
+
+  return (    
+    <div>
+      Ask Me Anything: <input onKeyPress={handleKeyPress} type="text" placeholder={placeholder} size="55"/>
+      <div className={`fixed width right-0 top-0 h-screen bg-white text-black transition-transform duration-200 ease-in-out transform ${isOpen ? 'translate-x-0' : 'translate-x-full'} z-50 overflow-y-auto w-full md:w-1/2 lg:w-1/3`}>
+        <div className="p-4">
+          <h2>...</h2>
```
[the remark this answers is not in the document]
michaelneale: yep - that was only there when didn't pop up over the header to pad it out.

site/src/components/ChatGPTSearch.jsx
```diff
@@ -0,0 +1,73 @@
...
+  }
+
+  const handleClose = () => {
+    setIsOpen(false);
+  }
+
+  return (    
+    <div>
+      Ask Me Anything: <input onKeyPress={handleKeyPress} type="text" placeholder={placeholder} size="55"/>
+      <div className={`fixed width right-0 top-0 h-screen bg-white text-black transition-transform duration-200 ease-in-out transform ${isOpen ? 'translate-x-0' : 'translate-x-full'} z-50 overflow-y-auto w-full md:w-1/2 lg:w-1/3`}>
```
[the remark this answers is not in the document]
michaelneale: oh LEGEND! that worked! finally!

angiejones:
how do we feel about a yellow bg and blue button? @kirahsapong @michaelneale 

image

michaelneale:
@angiejones yep - agree would look better. Need to work out how to make code blocks pop in that yellow scenario? ideas?

michaelneale:
@angiejones did that style

angiejones:
need to find somewhere else to put the prompt field. I'd like it to be on every page, so maybe a little search logo in the header that opens the side panel up? then you can type your query there and get the results in the same place

michaelneale:
@angiejones yes agree. Not sure how to get a component attatched on the header. Maybe @dayhaysoos can suggest some ideas?

dayhaysoos:
> @angiejones yes agree. Not sure how to get a component attatched on the header. Maybe @dayhaysoos can suggest some ideas?

So the header/navbar is auto generated based on configurations within the docusaurus-config.

This should be enough to put you in the right direction:

https://docusaurus.io/docs/api/themes/configuration#navbar

If you add more objects to the `items` array, you'll see that this item will be added to the navbar.

There's even a configuration for a search bar (maybe we can use this?)

https://docusaurus.io/docs/api/themes/configuration#navbar-search


If none of that seems feasible, I swear there's a way to use our own custom component in the navbar but I can't seem to find that in the documentation. Will do some research tomorrow if y'all don't find it. 

Lmk if you have any questions 🫡

michaelneale:
@dayhaysoos unfortunately couldn't find a way to put a custom component in navbar - they are only that fixed set of items or html from that list, I wasn't able to make it work :(

michaelneale:
ok trying this out - moving to the top - see how that feels?

site/src/content/home/web5-cta.js
```diff
@@ -1,34 +1,39 @@
...
+import React, { useState } from 'react';
 import HeroCard from '../../components/HeroCard';
 import Button from '../../components/Button';
 import Divider from '../../components/Divider';
 
 function Web5CTA() {
+
   return (
     <div>
       <h1>Web5 is open to build.</h1>
```
angiejones, line 10: for some reason, this isn't displaying as an H1

image

dayhaysoos: Might have something to do with the MDX Provider, I can take a look later

dayhaysoos:
> @dayhaysoos unfortunately couldn't find a way to put a custom component in navbar - they are only that fixed set of items or html from that list, I wasn't able to make it work :(

Another option would be to "swizzle" the nav bar component. This is something I've wanted to avoid but I'd be okay with swizzling so long as everything stays the same under the theme folder and the search bar is the only addition.

My concern with swizzling would be the long term affect of not being able to get updates from future versions of docusaurus. Not sure how much better the navbar can get but it's a trade-off I'd be willing to consider if that would make this integration much easier.

site/src/content/home/web5-cta.js
michaelneale: yeah I just removed it

michaelneale:
@dayhaysoos yeah swizzling doesn't seem right thing for that reason. I think we are close to something here? The text box needs more space than the header has on its own anyway don't you think?

angiejones:
what do yall think about just adding a Search icon to the header? Upon click, the side drawer opens and the prompt text field is in there. something like this...

image

michaelneale:
@angiejones I like that (need to work out how to make a link trigger that but should be doable).

michaelneale:
so question for @dayhaysoos: is it possible to put a link in the top bar that opens a page that is just the slide over (and no content change?) - that part of Docusoaurus goes beyond my react foo (and don't want to swizzle it as that would hurt maintainability).

dayhaysoos:
> so question for @dayhaysoos: is it possible to put a link in the top bar that opens a page that is just the slide over (and no content change?) - that part of Docusoaurus goes beyond my react foo (and don't want to swizzle it as that would hurt maintainability).

I think that's doable!

As a quick way to "try it on" maybe we can do something like this in the docusaurus config:

```json
 items: [
        {
          to: '#search',
          label: 'Search',
          position: 'right',
          className: Maybe a button style?
        },
      ],
```

I'm using "#search" so that we don't change pages, and we can have some kinda listener that triggers the search UI based on the URL having a hash at the end. Just spitballing, we can prob come up with something a little better from here.

michaelneale:
@dayhaysoos nice - yeah I can try that.

michaelneale:
image

this works.

michaelneale:
image

michaelneale:
ok @angiejones how am I doing?

dayhaysoos:
> image
> this works.

🔥🔥🔥

Yeah we should definitely do something to make the search link stand out. Not sure if it should look like a button or something as small as having a different border color or something to stand out. But this looks sick!

angiejones:
loving it!

angiejones:
why does the API take so long to return the answer? it's taking about 15-20 seconds

michaelneale:
@angiejones yes that is to be expected for chatgpt api calls. What i can do one day (if I get time) is to use the streaming version of the api so it writes stuff as it goes like the plugin for the chatgpt gui does

angiejones:
> @angiejones yes that is to be expected for chatgpt api calls. What i can do one day (if I get time) is to use the streaming version of the api so it writes stuff as it goes like the plugin for the chatgpt gui does

ohhh now it makes sense why they have the typewriter effect lol

michaelneale:
does it look better now @angiejones ? 
image


I don't think I will get time to do the typewriter type UX that chatgpt has: https://til.simonwillison.net/gpt3/python-chatgpt-streaming-api - but it is possible at some point (it means using some kind of SSE/websocket stuff). Perhaps as a subsequent PR? 

Well firstly, does the typewriter-effect (I like that name!) help with the UX - that it is doing something?

angiejones:
> does it look better now @angiejones ? image
> 
> I don't think I will get time to do the typewriter type UX that chatgpt has: https://til.simonwillison.net/gpt3/python-chatgpt-streaming-api - but it is possible at some point (it means using some kind of SSE/websocket stuff). Perhaps as a subsequent PR?
> 
> Well firstly, does the typewriter-effect (I like that name!) help with the UX - that it is doing something?

absolutely it does. since you mentioned it, i've been paying attention when I use chatgpt and it makes a huge difference just knowing that *something* is happening

michaelneale:
@angiejones ok try it again now - was more work than it appears but it does the typewriter effect now and looks much better (I had to learn how to use server sent events again)

angiejones (approved):

PR #602, opened 2023-07-14
Update records.create to include published and datePublished attributes 

bobbilee19:


site/api/web5-js/dwn/records.mdx
```diff
@@ -84,6 +84,22 @@ Creates a new record and stores it in the user's local DWN, remote DWNs, or anot
...
+                                        field: 'published',
+                                        value: 'When records are created, they are private by default. To make a record publicly available, published must be set to true upon creation',
+                                    },
+                                },
+                                {
+                                    data: {
+                                        type: 'string',
+                                        isOptional: true,
+                                        field: 'datePublished',
+                                        value: 'The date the record was published. When specifying to publish a record without providing the datePublished attribute, the record will be published now',
```
angiejones, line 101: Need to specify the format the timestamp must be in. 

> The date to make the record public. It should be in ISO 8601 ('YYYY-MM-DDThh:mm:ss.ssssssZ') format. Note that if the datePublished field is omitted and published is set to true, the record will be published upon creation.

angiejones (changes requested):

site/api/web5-js/dwn/records.mdx
angiejones: I've fixed this

angiejones (approved):

PR #618, opened 2023-07-27
Add styling to search widget 

kirahsapong:
Adds styling to search widget to add 
- list numbering
- list bullets
- paragraph spacing
- more prominent label heading
Screenshot 2023-07-27 at 3 55 40 PM

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2,5 +2,1772 @@
   "name": "developer.tbd.website",
   "lockfileVersion": 3,
   "requires": true,
-  "packages": {}
+  "packages": {
+    "": {
+      "hasInstallScript": true,
+      "dependencies": {
+        "@tbd54566975/web5": "0.7.11"
+      }
+    },
+    "node_modules/@assemblyscript/loader": {
+      "version": "0.9.4",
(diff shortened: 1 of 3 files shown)


ALRubinger (changes requested):
Thanks @kirahsapong! Looking awesome already!

Could we make it support longer inputs so the user can read the full thing? Maybe textbox that grows as input does?

image

ALRubinger (approved):
DOPE.

LGTM.

michaelneale:
@ALRubinger how do I get access to the vercel preview app?

michaelneale (changes requested):
when I press enter - it adds a new line and then runs the chat, which looks a bit odd: 

image could it not add the newline on Enter?

ALRubinger:
@michaelneale - we're still on Netlify; Vercel is a temp backup we can roll over to if necessary. @kirahsapong's deploy preview is here from the Netlify bot - https://github.com/TBD54566975/developer.tbd.website/pull/618#issuecomment-1654708322

michaelneale (approved):

PR #629, opened 2023-07-31
Issue 592/tests v2

dayhaysoos:
This is a second attempt at this issue:

https://github.com/TBD54566975/developer.tbd.website/pull/617

Both tests and building docusaurus work fine. Had to set a flag for proper browser support using vitest


site/src/util/web5.js
```diff
@@ -0,0 +1,17 @@
+import { Web5 } from '@tbd54566975/web5/browser';
+
+export async function didCreate() {
```
angiejones, line 3: should this just return the response instead of destructuring it? You already pull `web5` and `aliceDid` out and assign them in the `beforeAll`.

So something like

```js
return await Web5.connect();
```

site/src/util/web5.js
angiejones, line 1: the name and filepath here confuse me. I wouldn't know that this was for testing purposes. Can we add something to the filename or directory structure to indicate the purpose of this file?

site/src/util/web5.js
```diff
@@ -0,0 +1,17 @@
...
+  const { did: aliceDid, web5 } = await Web5.connect();
+  return { web5, aliceDid };
+}
+
+export async function dwnWriteTextRecord(web5, textData) {
```
angiejones, line 8: rename to `createTextRecord`

site/src/util/web5.js
```diff
@@ -0,0 +1,17 @@
...
+  const { record } = await web5.dwn.records.create({
+    data: textData,
+    message: {
+      dataFormat: 'text/plain',
+    },
+  });
+
+  return record;
```
angiejones, line 16: should return the entire response so that the status code can also be verified

ALRubinger (approved):
LGTM! "YEER" 🤣

```
pnpm test                       in zsh at 14:30:31

> @ test /Users/alr/git/TBD54566975/developer.tbd.website
> pnpm vitest --config vite.config.ts


 DEV  v0.33.0 /Users/alr/git/TBD54566975/developer.tbd.website
      Browser runner started at http://localhost:63315

 ✓ site/__tests__/quickstart.test.js (5) 2527ms
   ✓ /site/tests/quickstart.test.js (5) 2526ms
     ✓ didCreate returns a decentralized ID
     ✓ dwnWriteTextRecord returns a record with aliceDid being the same value as the author attribute 666ms
     ✓ recordResult returns a record with the textInput being the same value as the data attribute
     ✓ recordResult successfully sends an updated textInput
     ✓ recordResult successfully deletes the record

 Test Files  1 passed (1)
      Tests  5 passed (5)
   Start at  14:31:51
   Duration  6.83s (transform 0ms, setup 0ms, collect 910ms, tests 2.53s, environment 0ms, prepare 0ms)
```

site/__tests__/quickstart.test.js
```diff
@@ -0,0 +1,55 @@
...
+    const record = await dwnWriteTextRecord(web5, textInput);
+    recordResult = record;
+    expect(record.author).toBe(aliceDid);
+  });
+
+  test('recordResult returns a record with the textInput being the same value as the data attribute', async () => {
+    const textResult = await recordResult.data.text();
+    expect(textResult).toBe(textInput);
+  });
+
```
angiejones, line 39: need a test that verifies the status code of create

site/__tests__/quickstart.test.js
```diff
@@ -0,0 +1,55 @@
...
+  test('recordResult successfully sends an updated textInput', async () => {
+    const updatedRecordResult = await recordResult.update({
+      data: updatedTextInput,
+    });
+
+    expect(updatedRecordResult.status.code).toBe(202);
```
angiejones, line 45: if this fails, does it do the next check? I want to know everything that's wrong when the tests run. I think you need to run it as [expect.soft()](https://vitest.dev/api/expect.html#soft) so that both expect statements are executed and we know if we have 2 problems on our hands

site/__tests__/quickstart.test.js
```diff
@@ -0,0 +1,55 @@
...
+    expect(updatedTextInput).toBe(await recordResult.data.text());
+  });
+
+  test('recordResult successfully deletes the record', async () => {
+    const deletedRecordResult = await recordResult.delete();
+    expect(deletedRecordResult.status.code).toBe(202);
```
angiejones, line 51: use expect.soft() when a test has multiple expect statements or break these into separate tests.

angiejones (changes requested):

site/src/util/web5.js
dayhaysoos: This isn't necessarily for testing, it's a file that separated the functions in the quickstart.js so that we would have the ability to test them.

They pretty much are helper functions for web5 within the context of docusaurus. Perhaps I can name them `web5-helpers`? What do you think?

site/src/util/web5.js
dayhaysoos: Updated, added `expect.soft()` to the status code test

dayhaysoos:
Hey @angiejones, every point should be resolved except for renaming the web5 file. lemme know if I happen to have missed something/

Lemme know what you think of my proposed change for the web5 file to `web5-helpers`

site/__tests__/quickstart.test.js
dayhaysoos: Just to answer your question, it'll fail and abort everything after. I've added the soft function to each test with the status codes on them.

site/src/util/web5.js
angiejones: or `code-snippets` or something that's distinct from the actual web5.js library

angiejones (approved):

PR #633, opened 2023-08-02
Draft: Issue 630: Widget Tests

danlourenco:
@dayhaysoos -- as I dug into this, I realized that some additional refactoring may be needed.  Because the JavaScript in `examples/web5-quickstart-widgets/index.js` is not co-located with the HTML (as it is in `docs/web5/quickstart.js`), there are test failures when event handlers/query selectors attempt to find the DOM elements on which they are operating.

I still need to do a little more digging, but it might make sense for this code that manipulates the DOM to live in the HTML in a script tag, and any pure functions (or functions that otherwise don't touch the DOM directly) can live in index.js.  

Would love your thoughts on what would make the most sense here.  Thanks!

### DIFF
--- a/examples/web5-quickstart-widgets/__tests__/widget.test.js
+++ b/examples/web5-quickstart-widgets/__tests__/widget.test.js
@@ -1,0 +1,42 @@
+import { test, beforeAll, expect, describe } from 'vitest';
+import { didCreate, createTextRecord } from '../index';
+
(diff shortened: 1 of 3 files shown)


dayhaysoos:
Hey @danlourenco, I agree with this approach. I tried messing around with the vite config and adding jsdom for browser APIs but I think that's going to be too much. We just want to test the Web5 stuff for now. You wanna give that a shot and see how it goes?

PR #640, opened 2023-08-04
Update todo-app-tutorial.mdx

acekyd:
Proposed changes:
- **Closed braces for open functions:** Users are expected to run the code at different steps in the tutorials, this would result in an error unless they catch that the braces aren't closed.
- **Updated `dwn.records.write` to `dwn.records.create`**: While `.write` [is valid](https://developer.tbd.website/api/web5-js/dwn/records#writerequest), `.create` is what is used in the code snippet referenced. Updated this to avoid confusions.


angiejones (approved):
thanks @acekyd! 🙏🏾

PR #649, opened 2023-08-10
Adding new blog post

EbonyLouis:
Adding new blog post on who really owns your social media handles and how DIDs are a solution


ALRubinger (changes requested):
Please review extra dep as I'm noting in line changes?

package.json
```diff
@@ -10,7 +10,8 @@
     "test": "pnpm vitest --config vite.config.ts"
   },
   "dependencies": {
-    "@tbd54566975/web5": "0.7.11"
+    "@tbd54566975/web5": "0.7.11",
+    "react-markdown": "^8.0.7"
```
ALRubinger, line 14: Should this new dependency be added for the new blog post? Seems off in this PR, no?

ALRubinger (changes requested):

package.json
EbonyLouis: when I was trying to get it to compile locally, It kept saying it couldn't resolve 'react-markdown' so I had to npm install it. I can remove it.

dayhaysoos: Pretty sure this is due to mic's changes for the ChatGPT plugin

angiejones (changes requested):

package.json
angiejones: so should this line be removed?

dayhaysoos: @angiejones she just recently pulled in changes from main so I think it's fine now.

ALRubinger: Yeah even if it's needed on the dev site, should be in the deps for the `site` workspace and not top-level. Sounds sorted now; thanks all.

ALRubinger: Rather, I guess it'll be sorted once this line is off the diffs on this PR :)

angiejones: confirming that this is already in main
https://github.com/TBD54566975/developer.tbd.website/blame/main/site/package.json#L38

angiejones (approved):

PR #654, opened 2023-08-11
Starting point for code snippet system

dayhaysoos:
In this PR, we are working towards a system that allows us to test the actual code snippets being displayed in the docs while also allowing these functions to tested by our test suite.

Changes made so far:

A `./site/code-snippets/` directory where we have 1:1 files for each page that has code snippets. This might be worth updating to reflect the same directory structure as the docs themselves to make it easier to navigate and set a convention. Right now we just have a `query-from-dwn` file for the initial pass. 

- `extract-snippets` function at the root level that gets ran by `node extract-snippets`. It looks at the functions within the `code-snippets` directory and maps them to a JSON file to be imported. 

This gets run before the following commands:

`pnpm start`
`pnpm build`
`pnpm test`
`pnpm test:watch`

A `<CodeSnippet />` component for displaying the functions in the `code-snippets` directory. You pass the code snippet function name to the Code Snippet component and it displays the function as written within the body of the files function.

This is still a WIP but would like to get feedback early.


site/src/util/code-snippets-map.json
angiejones, line 1: what's the purpose of the json map?

dayhaysoos: @angiejones this gets generated by the script so that we can import the content and wrap it into the <CodeSnippet /> component. I'd like to demo this for you when you get a chance, I think it's better to see how it all comes together via screen share.

angiejones (approved):

PR #658, opened 2023-08-12
Suggested Prereq Section Design Changes

bobbilee19:
I've already done this to the schema guide in my new PR -- so in this PR im going to edit a few places where I think this will look great too!

OLD way of doing things...
Screenshot 2023-08-11 at 9 28 06 PM

SUGGESTED new way of doing things...
Screenshot 2023-08-11 at 9 27 48 PM


site/docs/ssi/create-did.md
angiejones, line 1: since we're repeating the same thing over and over, I suggest making this a component and just embedding it in these guides. That way if we ever need to make any changes to this, we will only need to do so in one place

angiejones (changes requested):

angiejones:
opened https://github.com/TBD54566975/developer.tbd.website/issues/661 to make this a component

PR #688, opened 2023-08-25
Updating docs to be in parity with web5 v.08

angiejones:


dayhaysoos (changes requested):

dayhaysoos:
It looks like `0.8.0` was released:

https://www.npmjs.com/package/@web5/api

I'm going to see how this works locally and update here on how it goes.

PR #738, opened 2023-09-21
Update Todo app tutorial plus Codesandbox deploys

acekyd:
- Following the upgrade to "0.8.1", using `@web5/api/browser` became necessary as discussed on [Discord](https://discord.com/channels/937858703112155166/969272658501976117/1153615427335897121)

- Also fixed a vue plugin error with vite config. 

---
  - https://app.asana.com/0/0/1205533071412996


examples/tutorials/todo-completed/src/App.vue
```diff
@@ -2,7 +2,7 @@
 import { onBeforeMount, ref, toRaw } from 'vue';
 import { PlusIcon as PlusIconMini } from '@heroicons/vue/solid';
 import { CheckCircleIcon, TrashIcon } from '@heroicons/vue/outline';
-import { Web5 } from "@web5/api";
+import { Web5 } from '@web5/api/browser';
```
angiejones, line 5: @leordev is this a workaround or something we should publish as "necessary" to build web5 apps?

leordev: We can either use the Web5 exposed by `/browser`, which is actually the only way to make it work with Docusaurus for now. Or, you can use the raw `@web5/api/browser` but add the required polyfills. 

I'm not too familiar with Vue, but if it uses vite, you can try to use the polyfills imported in this example: https://github.com/TBD54566975/bundler-bonanza/blob/main/vite/vite.config.js#L12

It should be way lighter than using the `/browser` export.

angiejones: this app should work standalone and not have anything to do with docusaurus.

leordev: Oh, I do see a vite.config.js file below, please use `@web5/api` and add the polyfill that I referenced above in your vite config. Just make sure you install the package dependency (`vite-plugin-node-stdlib-browser`) as well!

acekyd: Also thinking about how we can communicate this in tutorials. Do we direct people to find specific polyfills for the frameworks they are using, or do we recommend adding `/browser` if the run into said error.

angiejones: I think we can make a troubleshooting guide for it?

acekyd:
Any updates to go ahead with merging this? Tests now passing. Troubleshooting guide for the `/browser` polyfills can come separately.

angiejones (approved):

angiejones:
> Any updates to go ahead with merging this? Tests now passing. Troubleshooting guide for the `/browser` polyfills can come separately.

all good to merge

Issue #735, opened 2023-09-20
Update Protocol docs to add `of` when `who` is not 'anyone'


angiejones:
`of` is only an optional property when `who` is not 'anyone'. Update guides (and code snippets) to reflect this

![image](https://github.com/TBD54566975/developer.tbd.website/assets/15972783/f5920617-b798-488f-a513-4ac8394d92fa)

Also need to add 'anyone' as a valid value for `who`.

```js
export type ProtocolActionRule = {
  /**
   * May be 'anyone' | 'author' | 'recipient'.
   * If `who` === 'anyone', then `of` must be omitted. Otherwise `of` must be present.
   */
  who?: string,

  /**
   * Protocol path.
   * Must be present if `who` === 'author' or 'recipient'
   */
  of?: string;

  /**
   * Action that the actor can perform.
   * May be 'read' | 'write'
   */
  can: string;
};
```

**Docs**
- https://developer.tbd.website/api/web5-js/dwn/protocols/#configurerequest
- https://developer.tbd.website/docs/web5/learn/protocols/

builtbybea:
Hey Team. I would love to pick this issue up

angiejones:
> Hey Team. I would love to pick this issue up

sorry forgot to close this one out. It was fixed allready

PR #741, opened 2023-09-22
clarified `of` and `anyone` in Protocols API guide

angiejones:
image

---
  - https://app.asana.com/0/0/1205533584577330


angiejones:
addresses #735

PR #756, opened 2023-09-27
add script to add web5 dependency to every package inside tutorials dir

dayhaysoos:


package.json
```diff
@@ -1,15 +1,15 @@
...
-    "start:widget": "pnpm --filter web5-quickstart-widgets start",
-    "build": "node extract-snippets && pnpm --filter site run build",
-    "clear": "pnpm --filter site run clear",
-    "test": "node extract-snippets && pnpm vitest run --config vite.config.ts",
-    "test:watch": "node extract-snippets && pnpm vitest --config vite.config.ts --watch"
-  },
+"scripts": {
+  "preinstall": "node propagateWeb5.js && npx only-allow pnpm",
+  "start": "node extract-snippets && pnpm --filter site run start",
+  "start:tutorial-todo-starter": "pnpm --filter todo-starter run dev",
```
angiejones, line 5: do we have to add lines here for each tutorial?

leordev (approved):
nice & handy script!! 🚀

acekyd (approved):
Looks good to me. Thanks Nick! 🔥

package.json
acekyd: No we don't since we now have CSB to run the examples. We still need to update the pnpm lock file I believe so tests also pass.

dayhaysoos: Hey @angiejones ! No, this preinstall script is only at the root level and runs every time we run `pnpm install`. No one needs to know about this script, dev rel doesn't even have to think about it.

acekyd: I assumed she was referencing line 5 with `tutorial-todo-starter`

PR #791, opened 2023-10-02
Feedback Widget Docs and Testing Instructions

leordev:
Add Feedback Widget basic docs and testing instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -131,6 +131,8 @@ And modify the parameters accordingly to the feature you want to play with! Chec
 - `FEEDBACK_WIDGET_API_URL`: enables feedback widget api url; [Feedback server app instructions](./apps/feedback-server/README.md)
 - `DOC_SEARCH_*`: enables the search bar powered by Algolia; [Algolia connecting instructions](https://docusaurus.io/docs/search#connecting-algolia)
 
+Remember, when these environment variables are not set these features are disabled by default.
+
 ### Note
 
 This repo made a switch from using `npm` to `pnpm` as the package manager in July 2023. If you have used prior versions of this repo built on your machine with `npm`, you may encounter build failures, like:
--- a/README.md
+++ b/README.md
@@ -116,6 +116,21 @@ The following scripts are available for running specific tasks within the monore
 - `pnpm clear`: Runs docusaurus clear
 - `pnpm build`: Runs the production build for the site so it can be hosted by a static web server. The completed build will be in the folder `site/build`.
 
+## Extra Features
+
+We have implemented extra features in the website that don't come turned on by default, as an example we have the Feedback Widget feature, which allows an user to upvote or downvote a docs page. 
+
+These features are enabled from environment variables set in the file .env, inside the site folder. To enable them, copy the example .env file:
+
+```sh
+cp site/.env.example site/.env
+```
+
+And modify the parameters accordingly to the feature you want to play with! Check their instructions below:
+
+- `FEEDBACK_WIDGET_API_URL`: enables feedback widget api url; [Feedback server app instructions](./apps/feedback-server/README.md)
+- `DOC_SEARCH_*`: enables the search bar powered by Algolia; [Algolia connecting instructions](https://docusaurus.io/docs/search#connecting-algolia)
+
 ### Note
 
 This repo made a switch from using `npm` to `pnpm` as the package manager in July 2023. If you have used prior versions of this repo built on your machine with `npm`, you may encounter build failures, like:
--- a/apps/feedback-server/.env.example
+++ b/apps/feedback-server/.env.example
@@ -1,4 +1,4 @@
-SERVER_SECRET=<server-secret: `openssl rand -base64 32`>
+SERVER_SECRET=yourSuperSecretForTheServer
 SERVER_ALLOWED_ORIGINS='http://localhost:3000,https://developer.tbd.website'
 
 # postgres db config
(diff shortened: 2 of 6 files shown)


apps/feedback-server/.env.example
```diff
@@ -1,4 +1,4 @@
-SERVER_SECRET=<server-secret: `openssl rand -base64 32`>
+SERVER_SECRET=yourSuperSecretForTheServer
```
ALRubinger, line 1: Should this be more descriptive? Might we have many servers with microservices like this? For instance, would Calendar go along here as well like `SECRET_CALENDAR_API_KEY` next to `SECRET_FEEDBACK_KEY` or something?

README.md
```diff
@@ -116,6 +116,21 @@ The following scripts are available for running specific tasks within the monore
...
+
+These features are enabled from environment variables set in the file .env, inside the site folder. To enable them, copy the example .env file:
+
+```sh
+cp site/.env.example site/.env
+```
+
+And modify the parameters accordingly to the feature you want to play with! Check their instructions below:
+
+- `FEEDBACK_WIDGET_API_URL`: enables feedback widget api url; [Feedback server app instructions](./apps/feedback-server/README.md)
```
ALRubinger, line 131: Default value is in the linked instructions? Or default is no value and the feature is disabled?

apps/feedback-server/.env.example
```diff
@@ -1,4 +1,4 @@
...
 SERVER_ALLOWED_ORIGINS='http://localhost:3000,https://developer.tbd.website'
```
ALRubinger, line 2: `SERVER_ALLOWED_ORIGINS` needs to account for running tests in a production-like environment like GitHub Actions or in Block infra?

apps/feedback-server/README.md
```diff
@@ -1,34 +1,93 @@
...
 ## Instructions for local execution and development
 
 ```sh
 cp .env.example .env
 
-# adjust env variables on the .env file
-# you will need a Google API Key with access to Spreadsheets
+# modify the .env variables as you prefer, there's no need for setting
+# a DB connection because we fallback to a in-memory mocked fake db for
+# tests purposes
```
ALRubinger, line 13: Good.

apps/feedback-server/README.md
```diff
@@ -1,34 +1,93 @@
...
 
 pnpm dev
 ```
 
 The server should be ready to support the local docusaurus Helpful/Not Helpful button clicks.
 
+By default, if you don't setup a PostgresDB connection in the config, it will be running with 
+an in-memory fake database (aka just an array 😝).
```
ALRubinger, line 21: Great fallback.

If we're really disciplined about it, the code abstracts the DB into an internal DB SPI, and we can have an impl of that SPI for the local in-memory array, and another impl for Postgres that's used when the env var is set. That way the feedback code itself always is the same, isolated from the storage mechanism. And the storage mechanism is its own SPI (Service Provider Interface, basically an internal API) that has 2 impls.

apps/feedback-server/README.md
```diff
@@ -1,34 +1,93 @@
...
-```
\ No newline at end of file
+# run the container with the default postgres from the root folder
+cd ../..
+docker compose up -d
+```
+
+### Querying
+
+After voting on a couple of pages, use your favorite DB Client (tip: SQLTools on VSCode) to
```
ALRubinger, line 70: Notes on how to connect to the DB if done locally? URL, user/pass?

ALRubinger (commented):
Review in comments for @leordev consideration

apps/feedback-server/.env.example
leordev: Depends... Are we planning to expand this service to have many utilities for the website? Or is it a microservice? (If you see the folder name it is the `apps/feedback-server`.)

README.md
leordev: Great point. I will add this clarification.

FEEDBACK_WIDGET_API_URL is by default empty and when it's empty we simply have the feature disabled!

apps/feedback-server/.env.example
leordev: If we do spin a staging environment yes, otherwise no, we don't want to call the production box with these tests to not mess with the real votes numbers. 

We could try to add filters to our queries to ignore these votes though, if you think it's valuable.

What is happening today is that when you are in the preview link and vote it seems like it goes through but we are actually rejecting these calls in the production box.

apps/feedback-server/README.md
leordev: That's exactly what I did! Check it out: https://github.com/TBD54566975/developer.tbd.website/blob/main/apps/feedback-server/db.js

We have two classes that implement the same methods PostgresDb and FakeDb. Thing is that we dont have interfaces in JS since everything is typeless (there's no way to create an official SPI). But yeah, they do share the same "interface", the FakeDb is an implementation of this DB SPI, but abstracting an array. (Unless I misunderstood)

apps/feedback-server/.env.example
ALRubinger: Ah great point on the folder namespacing - yes that makes sense as-is then

apps/feedback-server/.env.example
ALRubinger: Awesome. Great plan and makes sense to keep as-is until we find something that needs more allowed origins.

apps/feedback-server/README.md
ALRubinger: Crushing it. Excellent design!

angiejones (approved):

PR #792, opened 2023-10-03
Remove console statement and add retry logic for FeedbackWidget 

Developerayo:

- [x] ♻️ Refactor
- [x] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 👷 Example Application
- [ ] 🧑‍💻 Code Snippet
- [ ] 🎨 Design
- [ ] 📖 Content
- [ ] 🧪 Tests
- [ ] 🔖 Release
- [ ] 🚩 Other

### First Commit
This PR removes the not needed `console.info` statement which appears in the console when a rating is clicked.

#### Second Commit
In case of a failure in feedback submission due to any reason like network unreliability, added a retry mechanism which will attempt to resubmit the feedback up to three times before throwing an error in the console if it keeps failing.

### DIFF
--- a/site/src/hooks/use-feedback-rating.js
+++ b/site/src/hooks/use-feedback-rating.js
@@ -20,17 +20,29 @@ export const useFeedbackRating = () => {
     }
   }, []);
 
-  const submitUserRating = async (rating) => {
+  const submitUserRating = async (rating, maxTries = 3) => {
     if (!csrfToken) {
       throw new Error("Can't send feedback without CSRF token");
     }
-
-    // TODO: we need to handle cases where the rating request fails
-    setUserFeedback(rating);
-
-    await postFeedbackRating(feedbackWidgetUrl, csrfToken, rating);
+  
+    for (let currentTry = 1; currentTry <= maxTries; currentTry++) {
+      try {
+        setUserFeedback('submitting');
+        await postFeedbackRating(feedbackWidgetUrl, csrfToken, rating);
+        setUserFeedback(rating);
+        break; 
+      } catch (error) {
+        console.error(`Attempt ${currentTry} failed:`, error);
+        setUserFeedback('failed');
+        
+        if (currentTry < maxTries) {
+          const waitTime = currentTry * 1000;
+          await new Promise(resolve => setTimeout(resolve, waitTime));
+        }
+      }
+    }
   };
-
+  
   return {
     isFeedbackRatingEnabled,
     userFeedback,
--- a/site/src/hooks/use-feedback-rating.js
+++ b/site/src/hooks/use-feedback-rating.js
@@ -1,4 +1,5 @@
 import { useState, useEffect } from 'react';
+import { wait } from '../util/feedback-retry-wait';
 import useDocusaurusContext from '@docusaurus/useDocusaurusContext';
 
 export const useFeedbackRating = () => {
@@ -25,8 +26,6 @@ export const useFeedbackRating = () => {
     setCsrfToken(freshToken);
     return freshToken;
   };
-
-  const wait = (duration) => new Promise(resolve => setTimeout(resolve, duration));
 
   const submitUserRating = async (rating, maxTries = 3) => {
     let currentToken = csrfToken;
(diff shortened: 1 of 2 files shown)


dayhaysoos (approved):

site/src/hooks/use-feedback-rating.js
```diff
@@ -20,17 +20,29 @@ export const useFeedbackRating = () => {
...
+        await postFeedbackRating(feedbackWidgetUrl, csrfToken, rating);
+        setUserFeedback(rating);
+        break; 
+      } catch (error) {
+        console.error(`Attempt ${currentTry} failed:`, error);
+        setUserFeedback('failed');
+        
+        if (currentTry < maxTries) {
+          const waitTime = currentTry * 1000;
+          await new Promise(resolve => setTimeout(resolve, waitTime));
```
leordev, line 40: Suggestion: here you can create a utility function `wait` to make the code more readable!

site/src/hooks/use-feedback-rating.js
```diff
@@ -20,17 +20,29 @@ export const useFeedbackRating = () => {
     }
   }, []);
 
-  const submitUserRating = async (rating) => {
+  const submitUserRating = async (rating, maxTries = 3) => {
```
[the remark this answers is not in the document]
Developerayo: Hm, you are right, It will fail in both scineros, do not have to test it, to bypass this, we can create a new function to get a new CSRF token if the initial rating submission fails, what do you think?

site/src/hooks/use-feedback-rating.js
Developerayo: 🦾

site/src/hooks/use-feedback-rating.js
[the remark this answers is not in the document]
leordev: Yeah, the retry should entail refreshing the CSRF token + retrying.

site/src/hooks/use-feedback-rating.js
```diff
@@ -20,17 +20,42 @@ export const useFeedbackRating = () => {
...
-    }
+  const refreshCsrfToken = async (feedbackWidgetUrl) => {
+    const freshToken = await getFeedbackCsrfToken(feedbackWidgetUrl);
+    setCsrfToken(freshToken);
+    return freshToken;
+  };
 
-    // TODO: we need to handle cases where the rating request fails
-    setUserFeedback(rating);
+  const wait = (duration) => new Promise(resolve => setTimeout(resolve, duration));
```
leordev, line 29: Please move this to the `/util` functions, so other components can reuse it when needed

leordev (approved):
Just a small ask and then LGTM!

Thank you @Developerayo ! 👏 🚀

leordev:
I tested this locally and I noticed the request was not retrying. It was because the `postFeedbackRating` function was not awaiting and throwing the error.

Can you please replace the current fetch with the following?

```ts
  const res = await fetch(`${feedbackWidgetUrl}/feedback`, {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
      'x-csrf-token': csrfToken,
    },
    credentials: 'include',
    body: JSON.stringify(requestBody),
  });

  if (!res.ok) {
    throw new Error('Failed to submit feedback rating');
  }
```

We can merge as soon as you do this! 👍

site/src/hooks/use-feedback-rating.js
```diff
@@ -92,4 +92,7 @@ const postFeedbackRating = async (feedbackWidgetUrl, csrfToken, rating) => {
     credentials: 'include',
     body: JSON.stringify(requestBody),
   });
-};
+
+  if (!res.ok) {
+    throw new Error('Failed to submit feedback rating');
+  }
```
leordev, line 98: There's a missing curly brace (`}`) at the end of the file. I've no idea how the build is passing. Can you please fix it?

leordev (changes requested):

site/src/hooks/use-feedback-rating.js
Developerayo: Oops! Just pasted it in, lol

leordev (approved):

PR #1024, opened 2023-11-21
Issue 992/book reviews test

dayhaysoos:
Instead of a browser test, we're handling the test with vitest because it's a node project.

Features:

- Added `pnpm test:node` script at root level and on related projects (book-reviews is the only one)
- updated it to export the results of web5.connect() so we can make sure the DID coming back is a string


examples/tutorials/book-reviews/book-reviews.js
```diff
@@ -1,18 +1,16 @@
...
 //Node 18 users: the following 3 lines are needed
-import { webcrypto } from "node:crypto";
+import { webcrypto } from 'node:crypto';
 // @ts-ignore
 if (!globalThis.crypto) globalThis.crypto = webcrypto;
 
-
-
-const { web5, did: userDid } = await Web5.connect();
+export const { web5, did: userDid } = await Web5.connect();
```
angiejones, line 8: this is an example for users. our test logic shouldn't be pushed on to them for their example app. the only test really needed for this project is to verify the console output

angiejones (changes requested):

dayhaysoos:
After talking with Angie about this test, we're going to put a pause on this example and try to think about a better way to test without making too many changes to the source itse.f.

PR #1080, opened 2023-12-16
added todo video to tutorial

angiejones:
@acekyd I embedded your new video right into the tutorial so that people can easily discover it. Should do this with your videos going forward 

image

---
  - https://app.asana.com/0/0/1206174656027185


acekyd (approved):

acekyd:
Thanks Angie! Looks great on mobile too ✨

PR #1073, opened 2023-12-14
introducing web5/dids

angiejones:
## What type of PR is this? (check all applicable)

- [x] 🧑‍💻 Code Snippet
- [x] 🧪 Tests

## Description

This PR removes the outdated approach to create DIDs and adds new examples showing how to use web5/dids to create DIDs with various DID methods

## Added code snippets?
- [x] 👍 yes
- [ ] 🙅 no, because they aren't needed

## Added tests?

- [x] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ ] 🙋 no, because I need help

## Added to documentation?

- [ ] 📜 readme
- [ ] 📜 contributing.md
- [x] 📓 general documentation
- [ ] 🙅 no documentation needed

PR #1092, opened 2023-12-19
Introduce @web5/dids

leordev:
Replaces #1073 

- [x] 🧑‍💻 Code Snippet
- [x] 🧪 Tests

## Description

This PR removes the outdated approach to create DIDs and adds new examples showing how to use web5/dids to create DIDs with various DID methods

## Added code snippets?
- [x] 👍 yes

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 readme
- [ ] 📜 contributing.md
- [x] 📓 general documentation


site/__tests__/web5/build/decentralized-identifiers/how-to-create-did.test.js
```diff
@@ -1,31 +1,33 @@
...
 
-let web5;
+describe('create-dids', () => {
 
-describe('how-to-create-did', () => {
-  beforeAll(async () => {
-    web5 = globalThis.web5;
-    did = globalThis.did;
+  //timing out
+  test.skip('createDidAutomatically returns a DID', async () => {
```
leordev, line 12: Is there a reason why you removed the mock?

I'd hope that the test was working, no?

The reason why its timing out now is because since you removed the mock its trying to connect to the same web5 DWN store twice in the same process (now that we are running node tests)... 

Since we already setup the web5 globally in the `setup-web5.js`, if we need to run snippets that are connecting to web5, we should mock it to return the global instantiated one (which is exactly what the mock was doing)

angiejones: re-added

site/__tests__/web5/build/decentralized-identifiers/how-to-create-did.test.js
```diff
@@ -22,10 +28,27 @@ describe('how-to-create-did', () => {
     });
   });
 
+
+  //timing out
```
angiejones, line 32:
```suggestion

```

angiejones (approved):

PR #1119, opened 2024-01-03
Explaining protocol url as key.

acekyd:

- [x] 📝 Documentation Update

## Description

This PR adds explanation that the protocol url serves as the key for the protocol definition and can enable interoperability across apps.

## Related Tickets & Documents

Resolves #

## Added code snippets?
- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

## Added tests?

- [ ] 👍 yes
- [x] 🙅 no, because they aren't needed

### No tests? Add a note

---
  - https://app.asana.com/0/0/1206259358258317


site/docs/web5/learn/protocols.md
```diff
@@ -123,7 +123,7 @@ const protocolDefinition = {
 }
 ```
 
-The value for `protocol` is a URI that represents the protocol being configured.
+The value for `protocol` is a URI that serves as the key of the protocol being configured. If two or more applications use the same protocol, identified by the URL, you can create common experiences across these apps in a more interoperable way.
```
angiejones, line 126:
```suggestion
The value for `protocol` is a URI that serves as the key of the protocol being configured. If two or more applications use the same protocol, identified by the URI, users can enjoy common experiences across these apps in a more interoperable way. Ideally, the protocol URI should lead to a JSON Schema document that validates the configurations, however, this is not validated by Web5.
```

angiejones (changes requested):

angiejones (approved):

PR #1135, opened 2024-01-07
Add 'native build' error to Troubleshooting

angiejones:


blackgirlbytes:
I have two comments based on the questions the hackathon participants asked us. One on setting global state of web5. Can add a little later today.

site/docs/web5/troubleshooting/common-errors.mdx
```diff
@@ -109,3 +109,42 @@ This error occurs when you attempt to sign a verifiable credential using the Str
...
 
 This error occurs when attempting to interact with a DWN that does not have the required protocol installed. To solve this, **ensure both parties have successfully [installed the protocol definition](/docs/web5/learn/protocols#installing-a-protocol)**.
+
+## Error: No native build was found
+
+```
+Error: No native build was found for platform=darwin arch=arm64 runtime=node abi=115 uv=1 armv=8 libc=glibc node=20.10.0 webpack=true
+```
+
+This error sometimes appears when deploying your application.
```
blackgirlbytes, line 119:
```suggestion
This error appears when deploying a Next.js application. 
```

blackgirlbytes: This problem only happens with Next.js apps. I think other frameworks have different issues/solutions

site/docs/web5/troubleshooting/common-errors.mdx
```diff
@@ -109,3 +109,42 @@ This error occurs when you attempt to sign a verifiable credential using the Str
...
+```
+Error: No native build was found for platform=darwin arch=arm64 runtime=node abi=115 uv=1 armv=8 libc=glibc node=20.10.0 webpack=true
+```
+
+This error sometimes appears when deploying your application. 
+
+To resolve, import like such:
+
+```js
+import { Web5 } from "@web5/api/browser";
```
blackgirlbytes, line 124:
```suggestion

```
it has to be inside the useEffect when deploying to Next.js. I think this only works if you just do npm run dev but not npm run build so maybe we remove this? I dont know

site/docs/web5/troubleshooting/common-errors.mdx
```diff
@@ -109,3 +109,42 @@ This error occurs when you attempt to sign a verifiable credential using the Str
...
+```
+
+If you run into additional errors, add the import to `useEffect()` instead:
+
```
blackgirlbytes, line 128: Going to add a suggestion because the Hackathon participant that was on Discord with us was saying he can't do this because he has to have web5 as a global variable, so I want to include having web5 as a variable

site/docs/web5/troubleshooting/common-errors.mdx
```diff
@@ -109,3 +109,42 @@ This error occurs when you attempt to sign a verifiable credential using the Str
...
+ useEffect(() => {
+
+    const initWeb5 = async () => {
+      // @ts-ignore
+      const { Web5 } = await import('@web5/api/browser');
+      
+      try {
+        const { web5, did } = await Web5.connect();
+
+        if (web5 && did) {
```
blackgirlbytes, line 139: maybe something like this . I don't know if it's too much but just showing where web5 get set globally..and then we're good

```js
export default function Index() {
    const [web5, setWeb5] = useState(null);
    const [myDid, setMyDid] = useState(null);

    useEffect(() => {
        const initWeb5 = async () => {
            // @ts-ignore
            const { Web5 } = await import("@web5/api");

            try {
                const { web5, did } = await Web5.connect();
                setWeb5(web5);
                setMyDid(did);

                if (web5 && did) {
                    console.log("Web5 initialized");
                }
            } catch (error) {
                console.error("Error initializing Web5:", error);
            }
        };

        initWeb5();
    }, []);
}

export default Index;
```

site/docs/web5/troubleshooting/common-errors.mdx
angiejones: oh i should move this to the Next.js troubleshooting guide then

angiejones: it's already there, but it doesn't have `/browser`. is this correct? https://developer.tbd.website/docs/web5/troubleshooting/next#error-no-native-build-was-found-for-xxx

blackgirlbytes: OHHHHH. I forgot this is already in the Next.js section of troubleshooting guide. yeah I'm wrong. it DOES need the `/browser` part. I've never needed that part so I keep forgetting that part.  But yeah I think we just need to change 

`await import('@web5/api');` to `await import('@web5/api/browser'); and we're good. My bad.

blackgirlbytes: I just removed the rest of my comments since this already exist in the Nextjs troubleshooting guide

PR #1140, opened 2024-01-09
Key Management Guide

chris-tbd:
## What type of PR is this? (check all applicable)

- [ ] ♻️ Refactor
- [ ] ✨ New Feature
- [ ] 🐛 Bug Fix
- [ x] 📝 Documentation Update
- [ ] 👷 Example Application
- [ ] 🧑‍💻 Code Snippet
- [ ] 🎨 Design
- [ ] 📖 Content
- [ ] 🧪 Tests
- [ ] 🔖 Release
- [ ] 🚩 Other

## Description

This PR adds a key management page under the build/dids vertical. It will add Kotlin code snippets.

## Related Tickets & Documents
N/A

## Added code snippets?
- [x ] 👍 yes
- [ ] 🙅 no, because they aren't needed

## Added tests?

- [ ] 👍 yes
- [ ] 🙅 no, because they aren't needed
- [ x] 🙋 no, because I need help

### No tests? Add a note

Need help with Kotlin code snippet testing

## Added to documentation?

- [ ] 📜 readme
- [ ] 📜 contributing.md
- [ x] 📓 general documentation
- [ ] 🙅 no documentation needed

PR #1144, opened 2024-01-11
Add Language Switcher support

chris-tbd:
# Completed

* [x] `<Shnip>` component added to abstract away most of the language components
* [x] Language-specific links supported with `<LanguageSwitchLink>`
* [x] Doc header is decoupled from language components
* [x] Language selection is persisted globally across pages
* [x] Support for Kotlin `<CodeSnippet>` (without tests)
* [x] Removed demo files: `shnip-demo.mdx`, `shnip-demo2.mdx`

# Remaining Work

* [ ] Refactor `<Shnip>` and `<CodeSnippet>` to take a `name` property for new shnips (#1177 

# Current Example of Usage

```md
import assignVcJwt from '!!raw-loader!@site/snippets/web5/build/verifiable-credentials/assignVcJwt.snippet.js'

<LanguageSwitcher languages="JavaScript, Kotlin"/>

# Hello World

Depending on the language you're programming in, you'll want to <LanguageSwitchLink text='link to different content' links='{"Kotlin": "https://kotlinlang.org/", "JavaScript": "https://www.javascript.com/"}'/>. Sometimes, you'll use the exact same link [for both languages](https://www.google.com/);

<Shnip
  snippets={[{ snippetContent: assignVcJwt, language: 'JavaScript' }]}
  // 'inlineSnippets' is how we'll hardcode
  inlineSnippets={[
    {
      code: `
            val signedVcJwt = "eyJ0eXAiOiJKV1QiLCJhbGciOiJFZERTQSIsImtp..."
      `,
      language: 'Kotlin',
    },
  ]}
/>
```

# Desired Example of Usage

```md
<LanguageSwitcher languages="JavaScript, Kotlin"/>

# Hello World

Depending on the language you're programming in, you'll want to <LanguageSwitchLink text='link to different content' links='{"Kotlin": "https://kotlinlang.org/", "JavaScript": "https://www.javascript.com/"}'/>. Sometimes, you'll use the exact same link [for both languages](https://www.google.com/);

<Shnip
  snippets={[
      { name: 'assignVcJwt', language: 'JavaScript' }]}
      { name: 'kt_assignVcJwt', language: 'Kotlin' }]}
  ]}
/>
```

# Preview
https://deploy-preview-1144--tbd-website-developer.netlify.app/docs/shnip-demo2

## Desktop

image

## Mobile

image


site/docs/web5/build/decentralized-identifiers/how-to-create-did.mdx
```diff
@@ -2,8 +2,6 @@
 title: Create a DID
 ---
 
-import CodeSnippet from '@site/src/components/CodeSnippet';
```
angiejones, line 5: this breaks the DID guide. Need to add this back

image

angiejones (changes requested):

site/src/components/LanguageSwitchBlock.jsx
```diff
@@ -0,0 +1,15 @@
+// CodeSnippet.js
+import React from 'react';
+import { useLanguage } from './LanguageContext';
+
+const LanguageSwitchBlock = ({ javascriptContent, kotlinContent }) => {
```
angiejones, line 5: what about the other languages?

site/docs/web5/build/decentralized-identifiers/key-management.mdx
angiejones, line 1: let's add the switcher support to this guide in a followup PR because I proposed several changes to the content of this guide in https://github.com/TBD54566975/developer.tbd.website/pull/1140

site/src/components/LanguageContext.jsx
angiejones, line 1: what do folks think about making a `language` folder under `components` since there are 6 components needed for this feature? feels it may be wise to have a grouping

angiejones (changes requested):

site/src/components/LanguageContext.jsx
blackgirlbytes: I like that idea!

chris-tbd: Making that change now

chris-tbd:
Ready for reviews again

michaelneale:
image
could we make it a bit more obvious - I had to hunt around for it (scrolled right passed it!) somehow?

angiejones:
> image 

> could we make it a bit more obvious - I had to hunt around for it (scrolled right passed it!) somehow?

agreed. needs some overall styling

chris-tbd:
Updated the switcher to be clearer per @angiejones and @michaelneale 
Screenshot 2024-01-18 at 10 42 18 AM

site/docs/hello-world.mdx
```diff
@@ -0,0 +1,66 @@
...
+
+
+
+This is a sample of some opening text we might use.
+
+## Getting Started
+Depending on the language you're programming in, you'll want to <LanguageSwitchLink text='link to different content' links='{"Kotlin": "https://kotlinlang.org/", "JavaScript": "https://www.javascript.com/"}'/>. Sometimes, you'll use the exact same link [for both languages](https://www.google.com/);
+
+Regardless, you can switch programming language using blocks like the one below.
+<LanguageSwitchBlock>
```
angiejones, line 23: The team talked about having a control above the code blocks so that you wouldn't have to scroll back up if you wanted to switch when you encountered a code snippet. How does that get added?

Many of us agreed we'd like this as tabs. Is that possible with this system?

ALRubinger:
@chris-tbd On our topic of "selecting the language", [Stripe has a cool approach](https://stripe.com/docs/expand/use-cases):
Selecting a new language in the tab is a global selection, and all other code blocks change to the language selected. Means they don't need a separate language selector in the navbar (real estate IMO we should be protective of :slightly_smiling_face: ) WDYT?

image (1)

site/docs/hello-world.mdx
```diff
@@ -0,0 +1,66 @@
...
+    <div language="Kotlin">
+    <p>Because sometimes you might want to write Kotlin code talking about Key Management:</p>
+        <CodeBlock language="kotlin">
```
angiejones, line 26: wondering if <CodeBlock> is a requirement here or can we continue to use the 3 ticks

site/src/components/language/LanguageSwitchHeader.jsx
angiejones, line 1: I don't quite understand the need for this component. Will every guide need to use this? Ca it be included in the LanguageSwitcher component?

site/src/components/language/LanguageSwitcher.jsx
```diff
@@ -0,0 +1,32 @@
...
+import { useSupportedLanguages } from './SupportedLanguagesContext';
+
+const LanguageSwitcher = () => {
+  const { changeLanguage, updateUrl } = useLanguage();
+  var currentLanguage = useLanguage();
+  const { languages } = useSupportedLanguages();
+
+  const handleLanguageChange = (e) => {
+    const newLanguage = e.target.value;
+    console.log(`changing language to ${newLanguage}`);
```
angiejones, line 13:
```suggestion

```

site/src/components/language/LanguageContext.jsx
```diff
@@ -0,0 +1,44 @@
...
+  const updateUrl = (language) => {
+    if (language.length > 0) {
+      console.log(`updating url path lang with ${language}`);
+      history.push(`?lang=${language}`)
+    }
+  }
+
+  const changeLanguage = (newLanguage) => {
+    setLanguage(newLanguage);
+    console.log(`set language to ${newLanguage}`);
```
angiejones, line 36:
```suggestion

```

site/src/components/language/LanguageContext.jsx
```diff
@@ -0,0 +1,44 @@
...
+  useEffect(() => {
+    const langParam = queryParams.get('lang');
+    if (langParam) {
+      setLanguage(langParam);
+    }
+  }, [queryParams]);
+
+  const updateUrl = (language) => {
+    if (language.length > 0) {
+      console.log(`updating url path lang with ${language}`);
```
angiejones, line 29:
```suggestion

```

site/src/components/language/LanguageSupport.jsx
```diff
@@ -0,0 +1,22 @@
...
+import { useSupportedLanguages } from './SupportedLanguagesContext';
+import { useLanguage } from './LanguageContext';
+
+const LanguageSupport = ({ languages }) => {
+  const languagesArray = languages.split(',').map(function(language) { return language.trim()});
+  const { changeLanguages } = useSupportedLanguages();
+  const { changeLanguage } = useLanguage();
+
+  useEffect(() => {
+    console.log(`Languages supported are ${languages}. Defaulting to ${languagesArray[0]}`);
```
angiejones, line 12:
```suggestion

```

ALRubinger:
> > @chris-tbd On our topic of "selecting the language", [Stripe has a cool approach](https://stripe.com/docs/expand/use-cases): Selecting a new language in the tab is a global selection, and all other code blocks change to the language selected. Means they don't need a separate language selector in the navbar (real estate IMO we should be protective of 🙂 ) WDYT?
> > image (1)
> 
> @ALRubinger Chris brought this up in the call and said that the content of the doc may change too dependeing on language (e.g. links to API Ref for more info)

Yup, and also the exceptional cases of "what if a language is selected globally but some blocks don't have that language implemented".

site/docs/hello-world.mdx
```diff
@@ -0,0 +1,66 @@
...
+{`
+fun initKeyManagement(env: string, didUri: string){
+    var keyManager: KeyManager;
+\n
```
angiejones, line 30: just tried it locally and when i removed this, the page crashed. the need for the escape characters here and on lines 27, 37, 48, etc make this really clunky. Remember these will be executable code snippets run by our tests. We can't (and shouldn't need to) have these extra requirements in order to view code

angiejones: ok this works:

```html
    <div language="JavaScript">
        <p>Because sometimes you might want to write JavaScript code talking about credential creation:</p>
        <CodeSnippet functionName='deleteRecordFromDid'/>
    </div> 
```

this is more in line with how we'd use it. haven't tried for Kotlin since I don't think we have CodeSnippet support for that language yet

site/src/components/language/LanguageSwitchHeader.jsx
dayhaysoos: The language switcher component is just the drop down (or possible tabs in the future) that handles the logic on what to present.

This component is needed to render the relative language based on the switcher logic. Every guide is going to need it.

dayhaysoos (commented):
I don't have any feedback that wasn't already mentioned. Will let you respond to Angie and ALR before approval

site/src/components/language/LanguageSwitchHeader.jsx
angiejones: I'm referring to the header

<LanguageSwitchHeader text="Language Switcher Hello World"/>

chris-tbd: The intention behind the `LangaugeSwitchHeader` is that it's a replacement for our normal `h1` headers:

```
/// This is how we currently write the header/title to our docs, which will show an h1
# Key Management

/// This is how we'd write it going forward with this new class, which will automatically add the language
/// Switcher to the title instead of us having to add it to the navbar, which can be difficult in docusaurus.
<LanguageSwitchHeader text="Key Management"/>

site/docs/hello-world.mdx
chris-tbd: `<CodeBlock>` is unfortunately a requirement here because of the way MDX and JSX are compiled into HTML - basically once we start writing MDX inside of a JSX component like `LanguageSwitchBlock`, we can revert back to JSX again. Ideally I'd like to have us write the snippets in separate files that have the appropriate extension (i.e. .js or .kt) and then we have a way to suck those into MDX and cleanly convert them to strings.

site/docs/hello-world.mdx
chris-tbd: This is possible, although it adds a layer of complexity because now the tabs/dropdowns/any switching mechanism both observes and modifies the language state. It's not impossible by any means, just finicky. Across the board we don't seem to have consensus on exactly how language switching should work from a UI perspective, specifically where the switching affordance should go and how many there should be.

If you're okay with it I would like to start without having the tabs and then we can add those in as a follow-up, I was hoping to loop @ALRubinger in as well to see how we'd like to lay things out. I agree with him that going Stripe-style with one global switcher is ideal rather than on code snippets, because we are changing both code *and* text/links. Since more content is just changing than code I believe tab switcher is not as good as a top-level switcher.

site/docs/hello-world.mdx
chris-tbd: I spoke with @dayhaysoos and we outlined what our code snippets will look like once his new system is in place:

Snippets would live under site/__test__/kotlin/kotlin-sample-snippet.kt
That file could contain one or many kotlin code snippets, and each .kt file would contain all the snippets for a corresponding guide. We'd then add this import at the top of the `MDX` file:

`import kotlinSnippet from '!!raw-loader!@site/snippets/kotlin/hellow-world.snippet.kt'`

Then finally the `CodeSnippet` that I wrote onto this Hello World page would be replaced with the following, but still display the exact same snippet that would then be easily readable and formattable in the .kt file:

`<CodeSnippet snippet={kotlinSnippet} language="kotlin">`

chris-tbd (commented):
Addressed comments

site/src/components/language/LanguageSwitchHeader.jsx
angiejones: yes I'm asking can you take this away. It's a bit overkill just to write an H1 header. Seems the compnent is just adding some styling.  Can we add that to a stylesheet or to the related components?

site/docs/hello-world.mdx
angiejones: why cant we just put a lang-specific class on the tab and dropdown elements and when something is selected, you make that class active on all Tabs and Dropdowns. would that work?

chris-tbd: ^ That would mean we're swizzling the Tabs component or rolling our own like we are in this PR. Swizzling forks us from main docusaurus development which poses some risks for updates breaking functionality but isn't out of the question.

site/src/components/language/LanguageSwitchHeader.jsx
chris-tbd: Would it be okay to just remove this component and rely on code blocks for language switching then @angiejones

EbonyLouis:
@angiejones I started working off of my own branch when I moved to Chris' branch I ran into a state issue I'll debug tomorrow. On my branch I only needed to change this line in `languageContext.jsx`:
`const [language, setLanguage] = useState('js');` to get the snippets to render, but you'll see to get it to work I had to change the tabs to `js` instead of `JavaScript` as well. 

- I added the ability to hardcode within the Shnip component  
- To add the custom admonition we need to upgrade to v3.1 based on the docs, I'll test locally tomorrow to see if this v upgrade will break anything on the site if not implementing it will be straightforward.

angiejones:
> @angiejones I started working off of my own branch when I moved to Chris' branch I ran into a state issue I'll debug tomorrow. On my branch I only needed to change this line in `languageContext.jsx`: `const [language, setLanguage] = useState('js');` to get the snippets to render, but you'll see to get it to work I had to change the tabs to `js` instead of `JavaScript` as well.
> 
> * I added the ability to hardcode within the Shnip component
> * To add the custom admonition we need to upgrade to v3.1 based on the docs, I'll test locally tomorrow to see if this v upgrade will break anything on the site if not implementing it will be straightforward.

@EbonyLouis ok sounds like a lot of work. If it looks like something that will take more than a day or 2, let's go with the <Shnip> component you came up with

EbonyLouis:
1. This is the[ test file](https://deploy-preview-1144--tbd-website-developer.netlify.app/docs/web5/build/verifiable-credentials/testing?lang=JavaScript) that is utilizing the new `Shnip` component. Changing versions caused a lot of errors, this is something I can work through but I don't think we should let it hold us up. 
2. Passed on the last issue to Chris to get a new pair of eyes on it I feel like its a super simple fix that I'm overlooking.

angiejones:
updated issue description with latest info

EbonyLouis:
@angiejones all set

angiejones (approved):

PR #1164, opened 2024-01-25
Enable non-Web5 tests

dayhaysoos:
This PR introduces a fix so that we can have tests with web5 connect and identity agent co-exist.

Need to update the Readme to reflect how things should be done.


site/__tests__/setup-web5.js
angiejones, line 1: looks ok, just double checking that if i dont call setupWeb5() or setupIdentityAgent(), should I be able to run tbDEX tests without either being invoked?

angiejones (approved):

site/__tests__/setup-web5.js
dayhaysoos: @angiejones I'm like 75% confident then you can run tdDEX tests without either of them. I did have an issue where tests were failing because neither setUpWeb5() or setUpWeb5() were being invoked but I fixed it by removing the agents from globalThis after clearing db.

If we run into any issues similar to that with tbdex I'm sure we'll be able to adapt quickly

angiejones: ok thanks!

angiejones (approved):

site/__tests__/setup-web5.js
```diff
@@ -29,4 +27,40 @@ beforeAll(async () => {
...
+
+    for (const db of dbs) {
+      console.log('clearing db');
+      await db.clear();
+      console.log('closing db');
+      await db.close();
+      console.log('closed');
+    }
+    delete globalThis.identityAgent;
+    delete globalThis.web5?.agent;
```
leordev, line 64: Suggestion: just do a whole web5 delete: `delete globalThis.web5` and also the did `delete globalThis.did`

leordev (approved):

PR #1179, opened 2024-01-31
Removed tbDEX->API Ref from sidebar

angiejones:
Removed tbDEX-> API Reference Guide from sidebar since it pointed to the JS version. We're now using the API Reference Guides sidebar item to show a page with multiple SDKs

### DIFF
--- a/site/docs/tbdex/api_ref.mdx
+++ b/site/docs/tbdex/api_ref.mdx
@@ -1,8 +1,0 @@
----
-sidebar_position: 1
-title: API Reference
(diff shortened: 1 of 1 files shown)


michaelneale:
https://d6cwdj-51423.csb.app/docs/api doesn't have tbdex though - has the others.

angiejones:
> https://d6cwdj-51423.csb.app/docs/api

@michaelneale it's here: https://developer.tbd.website/docs/api

PR #1191, opened 2024-02-01
enhancing inline code snippets

blackgirlbytes:

- [ ] ♻️ Refactor
- [ ] ✨ New Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation Update
- [ ] 👷 Example Application
- [x] 🧑‍💻 Code Snippet
- [ ] 🎨 Design
- [ ] 📖 Content
- [ ] 🧪 Tests
- [ ] 🔖 Release
- [ ] 🚩 Other

## Description

This PR enhances the Shnip component and enables it to have the following:
###  Different language code snippet than what's specified in the tab. 
Here we are saying you're writing JavaScript, so we're assuming the code snippet is JS, but it's actually bash

Screenshot 2024-02-01 at 12 54 10 PM

But the new change enables this
Screenshot 2024-02-01 at 12 59 06 PM

### Line breaks
For inline code snippets, MDX would freak out if it saw a line break, so you had to do this
Screenshot 2024-02-01 at 12 54 17 PM

But now you can pass in the number at which you want to break the line at using `breakLineAt:[5]`
Screenshot 2024-02-01 at 12 59 00 PM

## Titles
We couldn't specify titles of the files at first
Screenshot 2024-02-01 at 2 24 58 PM
Screenshot 2024-02-01 at 2 25 05 PM

but now we can
Screenshot 2024-02-01 at 2 25 35 PM

Please review to make sure I didn't break any fundamental functionality that I might've missed.

### DIFF
--- a/site/src/pages/open-source/contributing.mdx
+++ b/site/src/pages/open-source/contributing.mdx
@@ -109,6 +109,69 @@ This is how you can add a generated code snippet and an inline hardcoded snippet
 />
 ```
 
+To display inline code in a language different from the tab's primary language, use `codeLanguage`. 
+If `codeLanguage` is unspecified, it defaults to `language`. 
+Example: For Bash in a JavaScript tab, set `language: 'JavaScript`', `codeLanguage: 'bash'`
+
+```js
+<Shnip
+  inlineSnippets={[
+    {
+      code: `
+npm install @web5/credentials
+npm install @web5/dids
+      `,
+      language: 'JavaScript', // The tab's context.
+      codeLanguage: 'bash', // Snippet's actual language
+    }
+  ]}
+/>
+```
+
+MDX and JSX struggle to parse line breaks.
+For controlled line breaks in inline snippets, use `breakLineAt` with line numbers in an array.
+Example: To break after lines 1 and 5, set breakLineAt: [1, 5]
+<Shnip
+  inlineSnippets={[
+    {
+      code: `
+import { VerifiableCredential, PresentationExchange } from "@web5/credentials";
+class SwiftiesFanClub {
+  constructor(level, legit) {
+    // indicates the fan's dedication level
+    this.level = level; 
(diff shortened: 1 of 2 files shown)


site/src/pages/open-source/contributing.mdx
```diff
@@ -109,6 +109,69 @@ This is how you can add a generated code snippet and an inline hardcoded snippet
...
+class SwiftiesFanClub {
+  constructor(level, legit) {
+    // indicates the fan's dedication level
+    this.level = level; 
+    // indicates if the fan is a genuine Swiftie
+    this.legit = legit; 
+  }
+}
+      `,
+      breakLineAt:[1, 5]
```
angiejones, line 148: i can see this getting out of sync if the snippet needs to be updated later. is this only needed for inlineSnippets? and what happens if you just literally add a new line in the code? it breaks?

angiejones (approved):

site/src/pages/open-source/contributing.mdx
blackgirlbytes: only needed for inline code snippets. it works for regular code snippets just fine. but let's say you wanted to do
```bash
npm install dids

npm install credentials
```

then it breaks..and it has something that do with how MDX and JSX interpret line breaks.  

Ace said that in a thread earlier today with us. Here's what he said: "Looked into the line break issue as well... turns out the compile error happens in the mdx before getting to shnip, so any functions in shnip to trim, parse etc won't fix it.
Might have to create a file to import vs hardcoding the complex snippets."

blackgirlbytes: im hoping we can move all the code snippets over from inline/hardcoded to actual code snippets, so there will be very few moments that we will need to add the line breaks. not really sure of how to solve it..first I was using regex. Thoughts @angiejones @acekyd ?

PR #1213, opened 2024-02-07
Presentation Definition kotlin updates

acekyd:

- [x] 📝 Documentation Update
- [x] 🧑‍💻 Code Snippet
- [x] 🧪 Tests

## Description

This PR adds Kotlin options for Presentation Definition

## Related Tickets & Documents

Resolves #

## Added code snippets?
- [ ] 👍 yes

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 readme
- [ ] 📜 contributing.md
- [x] 📓 general documentation

## [optional] What gif best describes this PR or how it makes you feel?

---
  - https://app.asana.com/0/0/1206530477402610

### DIFF
--- a/site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationDefinitionTest.kt
+++ b/site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationDefinitionTest.kt
@@ -15,7 +15,7 @@ import com.fasterxml.jackson.module.kotlin.registerKotlinModule
  */
 internal class PresentationDefinitionTest {
 
-val ObjectMapper = ObjectMapper()
+val objectMapper = ObjectMapper()
 val pd = PresentationDefinitionV2(
         id = "PD_JobApplication_123456",
         name = "Credentials Verification for Ethical Hacker Job Application",
@@ -101,3 +101,4 @@ val pd = PresentationDefinitionV2(
     // :snippet-end:
     }
   }
+}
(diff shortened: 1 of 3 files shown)


blackgirlbytes:
Looks pretty good to me! I only had one question which is: Do you think it would be worth it to add a note that validateDefinition will not return the below..instead it will just throw an error? 
```
 [
  Checked {
    tag: 'root.presentation_definition',
    status: 'error',
    message: 'input descriptor group should match the from in submission requirements.'
  }
]
```

Could be me doing too much, but I know the fact that this didn't have a result was a point of confusion for some folks

site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationDefinitionTest.kt
```diff
@@ -0,0 +1,104 @@
...
+
+import web5.sdk.credentials.PresentationExchange
+import web5.sdk.credentials.model.*
+import java.io.File
+import com.fasterxml.jackson.annotation.JsonInclude
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
+
+/**
+ * Tests backing the Create a DID Guide
```
angiejones, line 14: update

angiejones (approved):

angiejones:
> Looks pretty good to me! I only had one question which is: Do you think it would be worth it to add a note that validateDefinition will not return the below..instead it will just throw an error?
> 
> ```
>  [
>   Checked {
>     tag: 'root.presentation_definition',
>     status: 'error',
>     message: 'input descriptor group should match the from in submission requirements.'
>   }
> ]
> ```
> 
> Could be me doing too much, but I know the fact that this didn't have a result was a point of confusion for some folks

I sorted this out

PR #1228, opened 2024-02-12
Ability to run JS tests individually

angiejones:
Added `pnpm test:js` which allows us to only run the JS tests. With this, we can also run an individual test, making it quicker to verify the one test you're working on: ```pnpm test:js getOfferings.test.js```

---
  - https://app.asana.com/0/0/1206571002797197


blackgirlbytes:
YAY. I've been wanting this

blackgirlbytes (approved):

Issue #1248, opened 2024-02-16
Trial TBD Artifactory as the single external Maven2 repo except for Maven Central


ALRubinger:
Configure the `testsuite-kotlin` to depend upon only one external Maven2 repository - TBD's Artifactory instance and public-facing "virtual" repository. This repo aggregates all repos we depend upon and caches results so they'll be available in the case one of the underlying remote repos goes down.

URL is: https://blockxyz.jfrog.io/artifactory/tbd-oss-public-maven2

ALRubinger:
In PR #1249 with notes for next steps

ALRubinger:
Done in 65b5f7a3d38973fe97b57308b56f19db04326b4b

PR #1249, opened 2024-02-16
Issue #1248: Remove 3rdparty Maven2 repos, add TBD Public OSS Artifactory

ALRubinger:
Acts as acceptance test for:

* https://github.com/TBD54566975/open-source-programs/issues/116

### DIFF
--- a/site/testsuites/testsuite-kotlin/pom.xml
+++ b/site/testsuites/testsuite-kotlin/pom.xml
@@ -185,40 +185,15 @@
 
     <repositories>
         <repository>
-            <id>jitpack</id>
-            <name>jitpack</name>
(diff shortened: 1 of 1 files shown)


ALRubinger:
Wow. that worked exactly as advertised.

It does default to pull all Maven Central stuff through our repo; this adds traffic if we're asking people to use our repo. Wondering if that informs we should pull Maven Central off this repo so that the world's consumers default to using Maven Central for their builds and save us the traffic. 

Letting this sit as a draft PR for a bit; I'm not sure yet how exactly we want to lay out the Artifactory setup, but this works and that's great. Will come back to this when we've more info and are confident in our Artifactory layout. Great that this works.

PR #1253, opened 2024-02-16
remove gh package and update pnpm

dayhaysoos:
This is for the licensing errors we're getting. Testing something with this PR

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -517,9 +517,6 @@ importers:
       dotenv:
         specifier: 16.3.1
         version: 16.3.1
-      gh:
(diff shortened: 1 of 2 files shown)


ALRubinger:
@dayhaysoos Nice one! So we can just remove the dependency and nothing breaks? We're not even using it?

If we're not using it, heck yeah, approve this PR and get the dep outta town.

angiejones (approved):

ALRubinger:
@dayhaysoos With @angiejones and my ✅ you're clear to merge.

ALRubinger (approved):
🚀

PR #1280, opened 2024-02-23
adding filter fields to kotlin pd

blackgirlbytes:

- [x] 📝 Documentation Update
- [x] 🧑‍💻 Code Snippet

## Description

This PR adds the filter fields to the Kotlin Presentation Definition.

## Added code snippets?
- [x] 👍 yes

## Added tests?

- [x] 👍 yes

## Added to documentation?

- [ ] 📜 readme
- [ ] 📜 contributing.md
- [ ] 📓 general documentation

## [optional] What gif best describes this PR or how it makes you feel?

---
  - https://app.asana.com/0/0/1206680703724606

### DIFF
--- a/site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationExchangeTest.kt
+++ b/site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationExchangeTest.kt
@@ -103,35 +103,63 @@ internal class PresentationExchangeTest {
   fun `getLoanAppPresentationDefinitionKt `() {
     // :snippet-start: getLoanAppPresentationDefinitionKt
     val presentationDefinition = PresentationDefinitionV2(
-        id = "presDefIdloanAppVerification123",
-        name = "Loan Application Employment Verification",
-        purpose = "To verify applicant’s employment, date of birth, and name",
-        inputDescriptors = listOf(
-            // Employment Verification
-            InputDescriptorV2(
-                id = "employmentVerification",
-                purpose = "Confirm current employment status",
-                constraints = ConstraintsV2(
-                    fields = listOf(FieldV2(path = listOf("$.vc.credentialSubject.employmentStatus")))
-                )
-            ),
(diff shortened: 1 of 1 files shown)


site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationExchangeTest.kt
```diff
@@ -4,47 +4,96 @@ import org.junit.jupiter.api.Test;
...
+    )
+
+    val nameAndDobVc = VerifiableCredential.create(
+        type = "NameAndDobCredential",
+        issuer = issuerDid.uri,
+        subject = aliceDid.uri,
+        data = NameAndDobCredential(name= "alice bob",dateOfBirth = "1990-10-01" )
+    )
+
+    val signedNameAndDobVcJwt = nameAndDobVc.sign(issuerDid)
```
blackgirlbytes, line 37: I initially got this signed JWT from the JS example, but I had to create a new JWT because when I run validate submission and filter the DOB like this , an error is thrown.
```
"type": "string",
  "format": "date"
  ```
  
  This happens because the original JWT has a DOB that looks like 10-01-1990, but I think the filter is expecting the date to be formatted in a different date format (like ISO 8601, YYYY-MM-DD), so I just created a new one with the expected format.

angiejones: does the new jwt work with the JS code? if so, let's change the JS one to this new one

angiejones (approved):

site/testsuites/testsuite-kotlin/src/test/kotlin/docs/web5/build/verifiablecredentials/PresentationExchangeTest.kt
blackgirlbytes: oh i just saw this after i merged it. okay I will make a new PR and use that

PR #1318, opened 2024-03-10
Fixing Swift Testsuite

angiejones:

- [x] 🐛 Bug Fix

## Description

- Added Swift tests to the `package.json` file
- Change guides for Wallet Required SDKs and DID creation to use inlineSnippets for dependencies
- Deleted  Swift dependencies in the test suite files

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
     "start:widget": "pnpm --filter web5-quickstart-widgets start",
     "build": "pnpm run shnip && pnpm --filter site run build",
     "clear": "pnpm --filter site run clear",
-    "test": "pnpm run shnip && pnpm vitest run --config vite.config.ts --no-threads && pnpm test:kotlin && pnpm test:browser",
+    "test": "pnpm run shnip && pnpm vitest run --config vite.config.ts --no-threads && pnpm test:kotlin && pnpm test:swift && pnpm test:browser",
     "test:watch": "pnpm run shnip && pnpm vitest --config vite.config.ts --watch",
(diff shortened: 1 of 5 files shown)


angiejones:
@kirahsapong can you look into [this error](https://github.com/TBD54566975/developer.tbd.website/actions/runs/8223509451/job/22486295192?pr=1318#step:10:14355) with web5.swift?

image

blackgirlbytes:
thanks for catching!

blackgirlbytes (approved):

package.json
```diff
@@ -13,14 +13,15 @@
     "start:widget": "pnpm --filter web5-quickstart-widgets start",
     "build": "pnpm run shnip && pnpm --filter site run build",
     "clear": "pnpm --filter site run clear",
-    "test": "pnpm run shnip && pnpm vitest run --config vite.config.ts --no-threads && pnpm test:kotlin && pnpm test:browser",
-    "test:watch": "pnpm run shnip && pnpm vitest --config vite.config.ts --watch",
+    "test": "pnpm test:js && pnpm test:kotlin && pnpm test:swift && pnpm test:browser",
+    "test:apps": "pnpm --filter feedback-server test run",
```
leordev, line 17: nit: I'd try to filter the apps folder, this way when we add new apps we are covered here!

ALRubinger: @leordev It's a good callout - should we do it that way or instead make the command `test:feedback-server`? Like...I doubt any one command is going to handle execution of every app, no?

PR #1349, opened 2024-03-22
fix for updating pom file dependencies

dayhaysoos:
This fixes an issue where Kotlin tests aren't running


blackgirlbytes (approved):
Screenshot 2024-03-22 at 2 29 45 PM
I ran the kotlin test and confirm that i see them running and passing. thanks Nick

angiejones (approved):
say what, now?!

PR #1357, opened 2024-03-26
Sdk version component

dayhaysoos:
This PR introduces a new component for surfacing dependencies in documentation. The goal is to be adaptive to the multiple language SDKs we provide while also correctly referring to the correct version. Only step left is adding how to use it in the contributing doc


site/src/components/Dependencies.jsx
```diff
@@ -0,0 +1,67 @@
...
+    } else if (language === 'maven') {
+      return <GradleDependencies dependencies={dependencies} sdks={SDK_VERSIONS[language]} />;
+    }
+  }
+
+  let Dependency;
+
+  switch (language) {
+    case 'javascript':
+      Dependency = <PackageJson sdkVersions={SDK_VERSIONS} dependencies={dependencies} />;
```
finn-tbd, line 59: This should return the `<PackageJson>` component instead of assigning it to a variable and returning the variable.

site/src/components/Dependencies.jsx
```diff
@@ -0,0 +1,67 @@
...
+// Main Dependencies component
+function Dependencies({ language, dependencies, display = 'command' }) {
+  const {
+    siteConfig: { customFields },
+  } = useDocusaurusContext();
+
+  const { SDK_VERSIONS } = customFields;
+
+  if (display === 'command') {
+    if (language === 'javascript') {
```
finn-tbd, line 48: should this be a switch statement?

site/src/components/Dependencies.jsx
```diff
@@ -0,0 +1,67 @@
...
+
+// Main Dependencies component
+function Dependencies({ language, dependencies, display = 'command' }) {
+  const {
+    siteConfig: { customFields },
+  } = useDocusaurusContext();
+
+  const { SDK_VERSIONS } = customFields;
+
+  if (display === 'command') {
```
finn-tbd, line 47: should this be a switch statement?

site/src/components/Dependencies.jsx
```diff
@@ -0,0 +1,67 @@
...
+      return <GradleDependencies dependencies={dependencies} sdks={SDK_VERSIONS[language]} />;
+    }
+  }
+
+  let Dependency;
+
+  switch (language) {
+    case 'javascript':
+      Dependency = <PackageJson sdkVersions={SDK_VERSIONS} dependencies={dependencies} />;
+      break;
```
finn-tbd, line 60: add kotlin (maven + gradle)

site/src/components/Dependencies.jsx
finn-tbd: alternatively: document other possible values for `display`

site/src/components/Dependencies.jsx
dayhaysoos: I think so far the only language we need this for is javascript, going to follow up with Dev Rel on this

site/src/components/Dependencies.jsx
dayhaysoos: From the feedback I got from dev rel, might make sense to rename to "install"

CONTRIBUTING.md
```diff
@@ -38,6 +38,69 @@ We are centralizing our depencies to a single json object at the root level name
 
 Whenever you'd like to update a dependency for a specific language, you update that version in the respective language attribute. When you run `pnpm install` these versions will be propagated throughout the repo in the right places.
 
+### <Dependencies />
+
+We havea Dependencies component that helps you display required dependencies for each SDK.
```
blackgirlbytes, line 43:
```suggestion
We have a Dependencies component that helps you display required dependencies for each SDK.
```

blackgirlbytes:
I did a first pass. It mostly seems good, and I hope you don't mind..I pushed a commit to import the Dependencies component to the MDXComponent file because I tried using it in an MDX file and it said the component needed to be imported. 

Another thing is...I'm unclear how to make this work with the language switcher. For example: in this file https://developer.tbd.website/docs/web5/build/verifiable-credentials/fan-club-vc#setting-the-stage we show how to install for JS and how to install for Kotlin and you can just switch back and forth with the Shnip component which is a wrapper for Language Switcher, but how would i do that for the Dependencies component?

Not sure if I am using it wrong. Feel free to correct me!

finn-tbd (approved):

blackgirlbytes (approved):

site/src/components/Dependencies.jsx
angiejones, line 1: can we rename these components so there's a clear indicator in how they differ? From a glance, I can't tell what `Dependencies` does vs `Dependency`. I'd need to read the code each time I want to grab this

angiejones (changes requested):

site/src/components/Dependencies.jsx
dayhaysoos: What do you think about the name `<TabbedDependencies />`?

angiejones (approved):

PR #1359, opened 2024-03-27
support swift with syncing dependencies

dayhaysoos:
This PR adds Swift to the centralized dependency object `sdk-versions.json`

### DIFF
--- a/propagateSdkVersions.js
+++ b/propagateSdkVersions.js
@@ -92,9 +92,6 @@ const updatePomXmlProperties = async (filePath, sdkVersions) => {
 };
 
 
-
-
-
 // Function to recursively find and update pom.xml files, excluding node_modules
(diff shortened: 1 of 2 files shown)


blackgirlbytes (approved):
tested per our conversation on Slack with an arbitrary branch name and it updated. I think we will only go by branch names, but if ever in the future we switch to version numbers..then we can update it to do something like: `exact: "0.0.4"`. This looks fine for now!
Screenshot 2024-03-28 at 10 53 38 AM

blackgirlbytes:
oh actually i meant to add this picture 
Screenshot 2024-03-28 at 11 03 40 AM

blackgirlbytes (commented):
Actually, I tried to put in real branch names and it keeps saying it couldn't find those. Just putting that on your radar..but..maybe that's out of scope for your PR.

angiejones (approved):

PR #1372, opened 2024-04-04
Swift sdk versions adjustment and pom file fix

dayhaysoos:
This PR updates how swift dependencies get added to the <Dependency /> components as well as fix how the pom files get updated.


blackgirlbytes (approved):

site/testsuites/testsuite-kotlin/pom.xml
```diff
@@ -1,238 +1,198 @@
...
+                <groupId>io.ktor</groupId>
+                <artifactId>ktor-server-test-host</artifactId>
+                <version>${version.io.ktor}</version>
+                <scope>test</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
     <dependencies>
-
-      <!-- tbDEX Dependency Management (BOM)-->
```
leordev, line 29: @dayhaysoos why did you remove and delete the way the file was structured with the comments? Maybe it was an IDE autosave mistake?

site/testsuites/testsuite-swift/Package.resolved
```diff
@@ -122,17 +122,17 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/TBD54566975/tbdex-swift.git",
       "state" : {
-        "branch" : "release/v1.0.0-rc",
-        "revision" : "dafa0aac574c276f6279045db81427917e7ac032"
+        "revision" : "741a2c6a9220b5b7822309ea4120618429b1fc89",
+        "version" : "1.0.0"
```
angiejones, line 126: not ready to merge this til that other PR is merged

angiejones (changes requested):

site/testsuites/testsuite-kotlin/pom.xml
leordev: Since I don't see your PR as a blocker because we are manually setting the Swift and Maven versions manually, I'd rather have you take your time and figure this issue out and then we can merge even after 1.0.0 release.

WDYT?

cc @ALRubinger

site/testsuites/testsuite-swift/Package.resolved
leordev: Which other PR? The one with the 1.0.0 release? I believe this PR is going to be merged into that "other PR" if I'm getting it right!

angiejones: can we revert this change now? Not sure if it revision is different

site/testsuites/testsuite-kotlin/pom.xml
ALRubinger: +1 @leordev

PR #1382, opened 2024-04-10
fix to include root lvl package.json

dayhaysoos:
Propagate sdk versions covers every package.json except for the root level. This PR addresses that.


blackgirlbytes (approved):
looking good on the JS and root level package json side side. I put in some random numbers and it worked. I didn't really test the Swift or Kotlin part, but I believe that was already working
Screenshot 2024-04-10 at 3 55 24 PM
Screenshot 2024-04-10 at 3 55 43 PM

leordev (approved):

angiejones (approved):

PR #1412, opened 2024-05-02
Updating Resolve DID guide

angiejones:
Updated Resolve a DID guide to include resolving DHT DIDs in Swift. Also renamed files and touched up wording to be in line with other guides.


EbonyLouis (approved):
Ah I didn't realize that swift now handles `dht`, thanks Angie

blackgirlbytes (approved):

PR #1429, opened 2024-05-17
adding `credentialSchema` KT

EbonyLouis:
---
  - https://app.asana.com/0/0/1207276480942542


blackgirlbytes (approved):
code wise it looks good to me! But I don't know a ton about KCC so take my review with a grain of salt

angiejones (approved):

PR #1436, opened 2024-05-28
remove toml file

dayhaysoos:
removing netlify toml to see if it still builds correctly.


ALRubinger (approved):
Fixes the build issue on the new dev site, doesn't break the existing one; confirmed on Netlify builds

PR #1439, opened 2024-05-28
add blog component

dayhaysoos:
Adding blog component, will be setting up a "guide" page too

Update: Will set up "Guide" page in another PR


site-new/src/components/BlogCard.tsx
```diff
@@ -0,0 +1,52 @@
+import clsx from "clsx";
+
+type BlogCardProps = {
+  date?: string;
+  author?: string;
+  title?: string;
+  tags?: string[];
+  image?: string | null;
+  description?: string;
+  size?: "large" | "small";
```
angiejones, line 10: what's this mean?

site-new/src/components/BlogCard.tsx
```diff
@@ -0,0 +1,52 @@
...
+};
+
+function BlogCard({
+  date = "date",
```
angiejones, line 14: can you change "date" to the actual date format that's needed so we'll always have that example conveniently there

site-new/src/components/BlogCard.tsx
```diff
@@ -0,0 +1,52 @@
...
+  author = "author",
+  title = "Title",
+  tags = [],
+  image = require("/static/img/test-image.png").default,
+  description = "Some description here",
+  size = "small",
+}: BlogCardProps) {
+  const cardSizeClass = size === "large" ? "max-w-[840px]" : "max-w-[400px]";
```
angiejones, line 22: based on the layout are there ever multiple large cards? I thought only the Featured post would be large

site-new/src/components/BlogCard.tsx
```diff
@@ -0,0 +1,52 @@
...
+  description?: string;
+  size?: "large" | "small";
+};
+
+function BlogCard({
+  date = "date",
+  author = "author",
+  title = "Title",
+  tags = [],
+  image = require("/static/img/test-image.png").default,
```
angiejones, line 18: instead of "test-image.png", can you use this to let us know the ideal size of the image? Something like:

`require("/static/img/1200x800.png").default,`

angiejones (changes requested):

site-new/src/components/BlogCard.tsx
dayhaysoos: There's a large version of the blog card post and a small version. I think I'ma make small the default but it's just an option for different sizes/styling.

site-new/src/components/BlogCard.tsx
dayhaysoos: I don't know if we'd ever have pages with multiple large cards but we have the option to use large or small whenever/where ever we want from this size value

site-new/src/components/BlogCard.tsx
dayhaysoos: I'm actually not sure about this, gonna look into it tho

site-new/src/components/BlogCard.tsx
angiejones: you think blog cards would be used outside the blog?

PR #1448, opened 2024-06-18
removing type `known Customer Credential` JS

EbonyLouis:
## Purpose 
Based on slack conversation, type: `Known Customer Credential` will break LD processors who consume our credentials. So this PR removes that.  

## TO-DO:
Kotlin requires a `value` for `type` to be passed, we need to update this. Temporarily using a workaround. A [PR](https://github.com/TBD54566975/web5-kt/pull/321) has been created addressing this issue. I will push with this for now and circle back when a new KT version has been released.


decentralgabe (approved):

site/docs/tbdex/issuer/kcc/kcc-issuer.mdx
angiejones, line 1: Should remove `type` here:

image

angiejones, line 1: one more spot under Step 5

image

decentralgabe: yes!

angiejones (approved):

PR #1451, opened 2024-06-26
Adding close.data.success to Wallet guide

angiejones:

- [X] 📝 Documentation Update

## Description

The Wallet guide on Placing an Order did not show how to determine if the transaction was successful. This PR adds that info.

## Added code snippets?
- [X] 👍 yes

## Added tests?

- [X] 👍 yes

---
  - https://app.asana.com/0/0/1207651243616779


EbonyLouis (approved):
LGTM 🚀

- Only thought is I wonder if it would be helpful to show the user an example of an unsuccessful close and what that reason might be. (totally up to you)

PR #1458, opened 2024-07-03
Update AWS configuration message based on selected language.

acekyd:
This pull request includes a change to the `site/docs/web5/build/decentralized-identifiers/key-management.mdx` file. The change updates the current AWS config message to show when JS is selected. 

PS: This is hidden for other languages as there are no AWS KMS interfaces implemented for Web5 Kotlin and Swift.

### DIFF
--- a/site/docs/web5/build/decentralized-identifiers/key-management.mdx
+++ b/site/docs/web5/build/decentralized-identifiers/key-management.mdx
@@ -79,5 +79,11 @@ Below is an example that uses `AwsKeyManager` for production environments, and `
     { snippetName: 'initializeKeyManagementSwift', language: 'Swift'}
 ]} />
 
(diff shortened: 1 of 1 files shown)


angiejones (changes requested):
When choosing Swift, it looks like a link but it's not clickable. There's some key management stuff in that SDK but I'm not sure which is which. May need to check with Kirah.

acekyd:
> When choosing Swift, it looks like a link but it's not clickable. There's some key management stuff in that SDK but I'm not sure which is which. May need to check with Kirah.

I'll follow up on this. Thanks @angiejones

EbonyLouis (approved):

blackgirlbytes (approved):

angiejones (approved):

PR #1463, opened 2024-07-09
Example json viewer - will close after

blackgirlbytes:
This is an example for coworker


blackgirlbytes:
this doesnt need to be merged. this is just a PR for reference

PR #1562, opened 2024-09-03
QA feedback

dayhaysoos:
Applies feedback from @tallneil on TextIconCard


dayhaysoos:
Merging this for now, will follow up for QA review later today

PR #1564, opened 2024-09-03
Add SDK links to project pages

angiejones:
This change adds SDK links to the Project pages for [tbDEX](https://deploy-preview-1564--tbd-website-developer.netlify.app/projects/tbdex/#sdk) and [Web5](https://deploy-preview-1564--tbd-website-developer.netlify.app/projects/web5/#sdk). 

There's 3 new components:

- SdkCard
- TbdexSdk
- Web5Sdk

This PR also includes some additional cosmetic changes, like making non-headers white so it's easier to read, and removing the Web5 actors section of the page.

---
  - https://app.asana.com/0/0/1208214282628240

### DIFF
--- a/site/src/components/Actors.jsx
+++ b/site/src/components/Actors.jsx
@@ -25,7 +25,7 @@ const Actors = ({ content }) => {
               </h2>
             </div>
             <p
-              className={`copy text-white flex ${
+              className={`copy primary-white flex ${
                 !content.isWeb5 ? 'ml-[4.375rem]' : ''
               } desktop:ml-0`}
             >
@@ -45,7 +45,7 @@ const Actors = ({ content }) => {
               </h2>
             </div>
             <p
-              className={`copy text-white flex ${
+              className={`copy primary-white flex ${
                 !content.isWeb5 ? 'ml-[4.375rem]' : ''
               } desktop:ml-0`}
             >
@@ -65,7 +65,7 @@ const Actors = ({ content }) => {
               </h2>
             </div>
             <p
-              className={`copy text-white flex ${
+              className={`copy primary-white flex ${
                 !content.isWeb5 ? 'ml-[4.375rem]' : ''
               } desktop:ml-0`}
             >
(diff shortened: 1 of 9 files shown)


site/src/components/Project.jsx
```diff
@@ -25,7 +25,7 @@ const Project = ({
             <Description />
           </div>
         ) : (
-          <p className="copy text-primary-yellow my-0">{description}</p>
+          <p className="copy text-white my-0">{description}</p>
```
blackgirlbytes, line 28: Screenshot 2024-09-04 at 9 41 34 AM
 i think the DWN section is still yellow. is this on purpose?

blackgirlbytes: a lot of the logic in this file is really confusing..before this PR. I wish that they were components because I don't really get the ternary logic.  Ternary logic that I'm referring to below: 
```js
      <div className="pb-9 tablet:pb-[67px] tablet:grow">
        {typeof description === 'function' ? (
          <div className="copy text-primary-yellow my-0">
            <Description />
          </div>
        ) : (
          <p className="copy text-white my-0">{description}</p>
        )}
      </div>
      ```

File: site/src/components/Actors.jsx [hunk 2]
```diff
@@ -25,7 +25,7 @@ const Actors = ({ content }) => {
               </h2>
             </div>
             <p
-              className={`copy text-primary-yellow flex ${
+              className={`copy text-white flex ${
```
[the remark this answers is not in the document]
angiejones: there's both. can use the primary one tho

image

site/src/components/Project.jsx
angiejones: hopefully this gets changed to a component with Nick's work

blackgirlbytes (approved):
looks perfect!

site/src/components/Actors.jsx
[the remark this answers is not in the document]
blackgirlbytes: kk, it's a nitpick + it's already changed, but i was trying to say that text-white is built in tailwind css and then we made our own css variable for white..cuz if that's not the case, then we can use text-yellow or text-black which also exist in tailwind instead of primary-yellow or primary-black

PR #1852, opened 2024-10-23
lowercase content contribution

blackgirlbytes:
I approved a hacktoberfest PR but I didn't realized the content type had to be lower cased in order for it to work. I updated it to lower case. I guess for future , we should allow it to be case insensitive and do like variable.toLowerCase() on it or something


EbonyLouis (approved):
🎉

PR #1871, opened 2024-11-04
Finish component guide

dayhaysoos:
Wrapped up most components in component guide


dayhaysoos:
Not sure about the dupe website being built. Assuming it's okay but going to look into it

